PR #2, opened 2020-12-09
Add bold, code, link input rules

hamvocke:
This PR introduces input rules for common mark types. It allows you to type commonmark notation in rich-text mode and trigger automated conversion to bold and emphasized text, links and inline code.

This PR supports:

* `**bold**` and `__bold__`
* `*em*` and `_em`
* `[link](https://stackoverflow.com)` (including link validation to avoid entering URLs you wouldn't be able to enter via our link tooltip)
*  `` `code` ``

The fundamental implementation is inspired by [this discussion](https://discuss.prosemirror.net/t/input-rules-for-wrapping-marks/537). It's relying on regular expressions and a bunch of conventions surrounding them: the regex needs to match a capture group, the first group is the text to be kept for the marked text. I tried to make the conventions and regexes a little easier to work with by providing documentation and a good bunch of unit tests. I've also added e2e tests.

**Demo**
Kapture 2021-06-17 at 11 11 18


aalear:
@hamvocke Any plans to do more work here in the near term? I'm dying without the inline code input rule in RT mode. 😅

Happy to help with this PR or extract the code rule parts into a new one.

hamvocke:
I feel you! This is my most anticipated feature, personally. Let me try to get those e2e tests in today. With that, the only remaining thing is link validation. I'll try to come up with a suggestion for that but that. I'll keep you posted!

b-kelly (approved):
Nice work! I did a little playing around with this locally and you covered what looks like 90% of the functionality. I did correct a small bug where it would attempt to insert marks in places that they weren't allowed.

Additionally, I came up with a new approach for unit testing these rules. The method is not terribly straightforward (and a bit hacky), but it works pretty well.

lgtm!

PR #3, opened 2021-01-15
Fix overly eager underscore escaping

hamvocke:
Our `markdown-serializer` uses custom escape logic when serializing text nodes entered in rich-text mode. This custom escaping is a little too eager and causes problems with URLs that contain underscores.

Take an URL like this entered in rich-text mode:

```
https://example.com/some_path_here
```

When switching over to markdown mode (or saving the document as markdown), this would get converted to:

```
https://example.com/some\_path\_here
```

which is problematic for URLs (and the autolinking we do) since it messes with the original URL.

This PR makes our custom escape logic more strict. I'm not sure if this is causing any regressions I didn't check yet, but our test suite is only raising shruggie as the single problem that might arise. With this tweak, you can still enter a shruggie in rich-text mode and convert to markdown and back again.

### DIFF
--- a/src/rich-text/markdown-serializer.ts
+++ b/src/rich-text/markdown-serializer.ts
@@ -132,7 +132,7 @@ const defaultMarkdownSerializerNodes: MarkdownSerializerNodes = {
         let escapedText = state.esc(node.text, startOfLine);
 
         // built in escape doesn't get all the cases TODO upstream!
-        escapedText = escapedText.replace(/\\_/g, "\\\\_");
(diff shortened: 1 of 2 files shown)


b-kelly (approved):
Updated the fix to work for all intratext `_` characters to match the markdown spec. Additionally, I added some extra tests for this issue.

b-kelly (approved):
Updated the fix to work for all intratext `_` characters to match the markdown spec. Additionally, I added some extra tests for this issue.

PR #5, opened 2021-01-20
Menu Dropdown for Headings

hamvocke:
Currently, our rich-text button for _headings_ only allows inserting level-1 headings. For other levels, you either have to use input-rules or switch over to markdown mode.

This PR turns the static _heading_ button into a dropdown menu that allows choosing between levels 1, 2 and 3.

Screenshot 2021-01-20 at 10 16 58

The underlying component is similar to the tables dropdown menu. I've had to extend the functionality a bit to allow highlighting heading levels if the current selection does already have that heading level applied.

This PR is mostly based on a PR we had open on the old repo but never managed to get in.


src/rich-text/commands.ts
```diff
@@ -558,20 +558,27 @@ export function insertLinkCommand(
  * Creates an `active` method that returns true of the current selection is/contained in the current block type
  * @param nodeType The type of the node to check for
```
b-kelly, line 559: nit: Add `@param` docs for added `attrs` param

src/rich-text/commands.ts
```diff
@@ -558,20 +558,27 @@ export function insertLinkCommand(
...
  */
-function nodeTypeActive(nodeType: NodeType) {
+function nodeTypeActive(
+    nodeType: NodeType,
+    attrs?: { [key: string]: unknown }
+) {
     return function (state: EditorState) {
         const { from, to } = state.selection;
         let isNodeType = false;
+        let hasAttrs = attrs ? false : true;
```
b-kelly, line 568: nit: There's a more "JS-y" way of doing this:

```suggestion
        let hasAttrs = !attrs;
```

Additionally, I'd rename this to `passesAttrsCheck` or something similar. `hasAttrs` runs a bit counter to what the code is actually doing, which is checking that all attributes passed are checked and match the attrs on the node.

src/rich-text/commands.ts
```diff
@@ -652,23 +654,15 @@ export const createMenu = (options: CommonViewOptions): Plugin =>
                 {
                     key: "insertImage",
                     command: insertImageCommand,
-                    dom: makeMenuIcon(
-                        "Image",
-                        "Image",
-                        "insert-image-btn"
-                    ),
+                    dom: makeMenuIcon("Image", "Image", "insert-image-btn"),
```
b-kelly, line 657: I have no idea why this line reformatted. It formats exactly the same locally as you have here, so... I'm calling it good.

No changes necessary.

src/shared/menu.ts
```diff
@@ -272,9 +272,10 @@ export function makeMenuDropdown(
     title: string,
     key: string,
     visible?: (state: EditorState) => boolean,
+    active?: (state: EditorState) => boolean,
```
b-kelly, line 275: nit: Add `@param` docs for new `active` param

b-kelly (approved):
lgtm! Excellent work. I tweaked the css a tad, but it looks good other than that. Usually, I'd require tests for all new features, but it looks like we don't have _any_ tests for menus, so we're not any worse off here. Thanks for picking this feature up

aaronshekey:
When comparing to other editors, I think we need to be more consistent. We should close the popover when a header level is chosen.

1. Select a node
2. Hit the header button
3. Choose a header
4. Popover closes

This is the behavior in Google Docs, for example. I think we should consider doing the same here.

hamvocke:
I could've sworn that I added a playwright test for this. That's what sent me off on this wild e2e fix ride. Maybe I've got a local commit sitting idle on my work laptop.

Maybe we could get some unit tests in place for dropdown commands since they're coming with a good bunch of "show and hide" logic. So far I bailed because there was no prior art but I'll look into this again.

aaronshekey:
I've hacked together the behavior I was talking about with both headers and tables. I think it's probably the right thing to do, though I'm debating including "Paragraph" as a node style in the header menu. Thoughts there, @b-kelly @hamvocke?

hamvocke:
Pushed the E2E test (I indeed forgot to push yesterday...) and worked on the missing documentation and suggestions.

@aaronshekey the paragraph idea is interesting and there are other editors out there doing this. If we added _paragraph_ to the dropdown, we need to think about highlighting the `aA` dropdown button whenever your current selection points to a  regular paragraph because that would be in line with the behavior we use for headings. I think that'd be a little distracting. What do you think?

aaronshekey:
> I think that'd be a little distracting. What do you think?

Ah, yes. Hmm. How _annoying_ would it be to only highlight the button when the node is a header?

Additional thought: I'm calling these Headers. Is that wrong? Are they "Headings"?

hamvocke:
> How annoying would it be to only highlight the button when the node is a header?

On second glance, that's probably already working. Looks like we're only highlighting the "heading" button whenever there's an actual "heading" node under the selection.

> Additional thought: I'm calling these Headers. Is that wrong? Are they "Headings"?

[HTML docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements) tend to call then _headings_ so that's what I'm sticking to.

src/shared/menu.ts
```diff
@@ -317,22 +326,32 @@ export function makeMenuDropdown(
...
-    key: string
+    key: string,
+    active?: (state: EditorState) => boolean,
+    cssClasses?: string[]
 ): MenuCommandEntry {
     const button = document.createElement("button");
     button.type = "button";
-    button.className = `s-btn s-btn__unset grid--cell ta-left px12 py4 h:bg-black-050 c-pointer js-editor-btn`;
     button.dataset.key = key;
     button.innerHTML = title;
```
customcommander, line 342: Apologies for the unsolicited feedback from a total stranger and admittedly I know very little about the rest of the infrastructure around this particular piece of code but...

If `title` is just a string is using `.innerHTML` (a known XSS sink) worth the risk here? Perhaps `.innerText` would be safer option?

hamvocke: No need to apologize. Great that you're looking out!

You're right, `title` as it is right now is only used as a string and we should switch to `innerText` instead. We've got `cssClasses` as a secondary mechanism to provide additional styling if we need to.

Personally, I don't see the risk of running into an XSS here. We're not passing user-generated data to the `title` parameter and are in full control what we're passing as part of the editor. The only way someone could exploit this is by sneaking a vulnerability into this codebase - which of course can happen, but `innerHTML` is the least of our concerns in that case :)

hamvocke:
How do we want to proceed? Add a `paragraph` section in the heading dropdown or go without it? Do we want to get this feature in just like that or should we switch to a more coordinated feature release any time soon?

PR #6, opened 2021-01-20
Fix broken End-to-End tests

hamvocke:
Our E2E tests have been broken in a pretty bad way. This PR sets out to make everything run smoothly again. It's looking a bit wild due to a huge reformatting of one of our `index.hml` pages (I wanted to finally bite the bullet here) but there's mostly a bunch of small changes:

1. `jest-playwright.config.js` needs to be something different from the `jest.config.js` used during e2e tests. The  former is for `jest-playwright`, the latter the proper `jest` config. This has gone south with moving the config to the `config/` directory. Before that, `jest-playwright` by convention assumes that a `jest-playwright.config.js` is present in the _root_ directory.
2. I updated `jest-playwright` while we're at it
3. I documented how to debug borked Playwright tests in our `Readme.md`
4. I removed all `js-editor-toggle-state` and `js-editor-toggle-label` classes from the toggles in the new "settings" menu. This was a nasty one. Our tests assumed that `js-editor-toggle-state` was reserved for the mode switcher. Since this class has been copy/pasted to other items, too, it tried to click on those when we instructed the e2e tests to switch the editor modes (which we basically do in every single test).
5. I added running our end to end tests as part of our GitHub Actions build


package-lock.json
```diff
@@ -51,7 +51,7 @@
                 "highlight.js": "*",
                 "html-webpack-plugin": "^4.5.1",
                 "jest": "^26.6.3",
-                "jest-playwright-preset": "^1.4.3",
+                "jest-playwright-preset": "^1.4.4",
```
b-kelly, line 54: I pulled locally and updated the `package-lock.json`. I think your version got a lot of changes due to a potential npm version mismatch. My output looks like this for reference:

```
> npm -v
7.3.0
> node -v
v15.5.0
```

site/index.html
```diff
@@ -22,15 +22,15 @@
             <ul class="s-sidebarwidget--content s-sidebarwidget__items btw0 p0">
                 <li class="grid ai-center s-sidebarwidget--item">
                     <div class="grid--cell s-editor-toggle">
-                        <input class="js-editor-toggle-state" id="js-toggle-dark" type="checkbox" />
-                        <label class="js-editor-toggle-label" for="js-toggle-dark"></label>
+                        <input id="js-toggle-dark" type="checkbox" />
```
b-kelly, line 25: Good call on removing these. I copy/pasted (BAD!) and forgot to do so.

b-kelly (approved):
lgtm! Thanks for taking this work on. It's been on my todo list, but didn't quite make it before you got to it.

PR #7, opened 2021-01-20
Link popover: Fix messed up margin when editing

hamvocke:
When editing a link in rich-text mode, the input field gets extra bottom margin. That's due to a rather generous CSS rule in our `custom-components.less` that's got a lot of potential to mess with _any_ of our custom node views that use a `div` element:

```less
&.s-prose div {
        margin-bottom: var(--s-prose-spacing);
```

Here's what it currently looks like:

Screenshot 2021-01-20 at 12 22 05

In this PR I'm adding an explicit `mb0` to solve the immediate issue. We should think about making that custom CSS rule a little more specific.

Here's what it looks like with the fix in place:

Screenshot 2021-01-20 at 12 24 06

### DIFF
--- a/src/rich-text/plugins/link-tooltip.ts
+++ b/src/rich-text/plugins/link-tooltip.ts
@@ -51,7 +51,7 @@ class LinkTooltip {
             class="wmx3 grid--cell fs-body1 fw-normal d-inline-block truncate ml8 mr4"
             target="_blank"
             rel="nofollow noreferrer">${this.href}</a>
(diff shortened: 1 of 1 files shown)


b-kelly (approved):
This is a good quick fix. I'm going to look into a more long-term fix for the suspect CSS rule. We have that rule because ProseMirror will sometimes wrap stuff in `div`s and that messes with Stacks's `s-prose` rules a bit.

PR #9, opened 2021-01-25
Remove GH pages workflow

aaronshekey:
Now that Netlify is hooked up, we can get rid of this action.


b-kelly (approved):
Awww... poor thing didn't even last a month. lgtm

PR #10, opened 2021-01-27
Remove smart quotes, ellipsis, emdash input rules

hamvocke:
[Feedback on meta](https://meta.stackexchange.com/a/360067/507720) shows that automatically converting quotes to smart quotes in rich-text mode doesn't spark joy.

The reasoning is sound. On a programming website quotes often occur in code samples. Converting them to smart quotes will break code samples and we all know how much people love copy/pasting code samples.

This PR removes the usage of a bunch of Prosemirror's input rules that could make developers' lives harder:

* converting quotes to smart quotes (both, single and double quotes)
* converting `...` to &hellip;
* converting `--` to —

All of the above are used in code in some form: `--` is a decrement operator, `...` is used as the spread operator in JS and quotes are... quotes.

We should not try to be smart about these things. If people are adamant using smart quotes or typographically correct characters, they'll most likely know how to enter them (e.g. by using their operating system's settings or using Alt codes).

These input rules haven't been a deliberately requested feature anyways. I remember adding them as we were still in the prototyping phase and I didn't think much about it. Seems like the made it all the way to production because that's what good prototypes do.

If we find out that people want to keep these as an opt-in option, we can still bring them back as a setting for the editor or special sites on the network. For now: Let's get rid of them.

### DIFF
--- a/src/rich-text/inputrules.ts
+++ b/src/rich-text/inputrules.ts
@@ -1,8 +1,5 @@
 import {
-    ellipsis,
-    emDash,
     inputRules,
-    smartQuotes,
     textblockTypeInputRule,
     wrappingInputRule,
 } from "prosemirror-inputrules";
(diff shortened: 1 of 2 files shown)


aaronshekey (approved):

b-kelly (approved):
Looks good! I added a test to main confirming that the smart quotes _happen_, so I could alter the test here to prove they _don't_ happen.

Thanks for picking this up.

PR #15, opened 2021-02-16
Exploration: Preview Pane for Markdown Mode

hamvocke:
## Background

Okay, this is a controversial one.

Our old editor used a read-only preview pane next to your markdown code. With Stacks-Editor we want to move away from that pattern in favor of having a two-way synchronization between rich-text and markdown mode.

There's been recurring feedback on the meta announcement that a read-only preview when writing markdown would continue to solve a bunch of problems and allow people to stick to a workflow they're used to (see [here](https://meta.stackexchange.com/a/360058/507720), [here](https://meta.stackexchange.com/a/360146/507720) and somewhat buried also [here](https://meta.stackexchange.com/a/360142/507720)).

I know that [every change is going to break someone's workflow](https://xkcd.com/1172/) and it's important to have a clear direction for Stacks-Editor. At the same time, I don't want to dismiss users' concerns and explore what the solution that quite a bunch of people are asking for could look and feel like. So in this PR I've created a (admittedly hacky) version of an expandable read-only preview for Markdown-mode.

## Why, why not?
I can see how a preview pane can tackle a couple of problems:

* Support for Mathjax, Chess notation, Guitar Tabs, and all the other site-specific plugins would be comparably trivial to build into a read-only preview mode. We'll have to build these plugins for rich-text mode in the future, but it'll take some time until we get there. For some users, this could prevent switching to Stacks-Editor because they need these plugins to be there
* People who have strong feelings around the "write markdown on top, see preview below" could continue to do so in Stacks-Editor, making the adoption much more seamless
* We could get rid of supporting our old editor sooner. A read-only pane would give us feature parity to the old editor with one big swoop.

Of course this comes with a bunch of drawbacks:
* this might be confusing for the user experience. You'd have a markdown mode, a read-only preview, a rich-text mode, and ultimately the properly rendered and saved post.
* there's a risk that the preview pane becomes a kludge we rely on instead of building more polished solutions to the problems we're facing
* we end up with the same "endless scrolling" usability issue we've had before

There's a lot of ivory-tower discussion we can have about the benefits and drawbacks but I value working code over theoretical discussions. In this fashion, this PR introduces a read-only preview that's visible in Markdown-mode.

## Features of this spike

* show a read-only preview of your Markdown post below the editor
* hide read-only preview by default, show via toggle. This way, we'd hide the feature in the default experience and let people who explicitly want to have the questionable scrolling experience activate it. We could hook this up to user preferences on Stack Overflow.
* trigger read-only preview synchronization _after_ a user stops typing to prevent multiple costly render operations
* pass preview pane element via the editor options - initialize preview only if a preview element is given

Please note that this is not meant to be production-ready code at all. It comes without tests, I've cut a lot of corners and glossed over obvious bugs for the sake of getting a prototype in place. I encourage people to not focus on the code in this PR but to play around with the prototype instead.

**Check out [the demo](https://deploy-preview-15--nifty-lalande-39c157.netlify.app/)**

### DIFF
--- a/site/index.html
+++ b/site/index.html
@@ -166,11 +166,27 @@ And finally, Stack Snippets:
     </div>
 
 <!-- end snippet -->
-        </textarea>
-        <h1>Stacks Editor</h1>
-        <h3>Showcasing our new editing experience for Stack Overflow</h3>
-        <div id="example-1"></div>
-    </div>
-</body>
-
+        </textarea
+            >
+            <h1>Stacks Editor</h1>
+            <h3>Showcasing our new editing experience for Stack Overflow</h3>
+            <div id="example-1"></div>
+            <button
+                class="s-btn js-preview-toggle"
+                data-controller="s-expandable-control"
+                data-s-expandable-control-toggle-class="is-selected"
+                aria-expanded="false"
+                aria-controls="js-preview-expander"
+            >
+                Show preview
+            </button>
+            <div class="s-expandable" id="js-preview-expander">
+                <div
+                    class="js-preview-pane mt12 p12 s-card s-prose d-none s-expandable--content"
+                >
+                    Preview will appear here
+                </div>
+            </div>
+        </div>
+    </body>
 </html>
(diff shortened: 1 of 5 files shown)


b-kelly:
See also #115

b-kelly:
Moving forward on this in #146, so closing this PR

Issue #65, opened 2021-04-13
No preview when adding image for the second time


WouterDeKort:
**Describe the bug**

Browsing for an image, cancelling the image upload and then browsing for the same image does not show the image preview.

**To Reproduce**
1. Select the Image icon
2. Click on 'Browse'
3. Select a local image
4. Click on 'Cancel'
5. Select the Image icon
6. Click on 'Browse'
7. Select the same image as in step 3

**Expected behavior**
I expect the image to be shown as a preview and be able to 

**Screenshots**

![Cancel-Add-Image-Bug](https://user-images.githubusercontent.com/1763043/114533703-79379500-9c4e-11eb-99ef-a367ef4381d4.gif)

**Desktop**
 - OS: Windows 10
 - Browser Chrome
 - Version 89.0.4389.114

b-kelly:
Fixed in #67

Issue #66, opened 2021-04-13
Having two editors on a page, always adds an image to the first editor.


WouterDeKort:
**Describe the bug**
Having two editors on a page, always adds an image to the first editor.

**To Reproduce**

Add a second editor to the sample page.

1. Go to 'Image' on editor 1
2. Go to 'Image' on editor 2
3. lick on 'Browse' and select an image on editor 2

**Expected behavior**
The selected image should preview in editor 2, not in editor 1.

**Screenshots**

![Two-editors-bug](https://user-images.githubusercontent.com/1763043/114534496-578add80-9c4f-11eb-919c-e6460870f8f6.gif)

Desktop 

OS: Windows 10
Browser Chrome
Version 89.0.4389.114

b-kelly:
Fixed in #67

PR #67, opened 2021-04-13
Fixed issue for two editors and cancelling an image upload

WouterDeKort:
This pull request fixes issues #65 and #66.

#65 is solved by clearing the `uploadField.value `when switching between visible <--> invisible 
#66 is solved by making the `uploadField.id `unique per editor

For #66, I've added a second editor to the test page.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1718,6 +1718,7 @@
                 "jest-resolve": "^26.6.2",
                 "jest-util": "^26.6.2",
                 "jest-worker": "^26.6.2",
+                "node-notifier": "^8.0.0",
                 "slash": "^3.0.0",
                 "source-map": "^0.6.0",
                 "string-length": "^4.0.1",
@@ -3577,6 +3578,7 @@
                 "anymatch": "^2.0.0",
                 "async-each": "^1.0.1",
                 "braces": "^2.3.2",
+                "fsevents": "^1.2.7",
                 "glob-parent": "^3.1.0",
                 "inherits": "^2.0.3",
                 "is-binary-path": "^1.0.0",
@@ -6057,6 +6059,7 @@
             "integrity": "sha512-GDhU9ntwuKyGXdZBUgTIe+vXnWj0fppUEtMDL0+idd5Sta8TGpHssn/eusA9mrPr9qNDym6SxAYZjNvCn/9RBg==",
             "dev": true,
             "dependencies": {
+                "@types/yauzl": "^2.9.1",
                 "debug": "^4.1.1",
                 "get-stream": "^5.1.0",
                 "yauzl": "^2.10.0"
@@ -8054,6 +8057,7 @@
                 "@types/node": "*",
                 "anymatch": "^3.0.3",
                 "fb-watchman": "^2.0.0",
+                "fsevents": "^2.1.2",
                 "graceful-fs": "^4.2.4",
                 "jest-regex-util": "^26.0.0",
                 "jest-serializer": "^26.6.2",
@@ -8641,7 +8645,14 @@
             "dev": true,
             "dependencies": {
                 "copy-anything": "^2.0.1",
+                "errno": "^0.1.1",
+                "graceful-fs": "^4.1.2",
+                "image-size": "~0.5.0",
+                "make-dir": "^2.1.0",
+                "mime": "^1.4.1",
+                "needle": "^2.5.2",
                 "parse-node-version": "^1.0.1",
+                "source-map": "~0.6.0",
                 "tslib": "^1.10.0"
             },
             "bin": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1718,7 +1718,6 @@
                 "jest-resolve": "^26.6.2",
                 "jest-util": "^26.6.2",
                 "jest-worker": "^26.6.2",
-                "node-notifier": "^8.0.0",
                 "slash": "^3.0.0",
                 "source-map": "^0.6.0",
                 "string-length": "^4.0.1",
@@ -3578,7 +3577,6 @@
                 "anymatch": "^2.0.0",
                 "async-each": "^1.0.1",
                 "braces": "^2.3.2",
-                "fsevents": "^1.2.7",
                 "glob-parent": "^3.1.0",
                 "inherits": "^2.0.3",
                 "is-binary-path": "^1.0.0",
@@ -6059,7 +6057,6 @@
             "integrity": "sha512-GDhU9ntwuKyGXdZBUgTIe+vXnWj0fppUEtMDL0+idd5Sta8TGpHssn/eusA9mrPr9qNDym6SxAYZjNvCn/9RBg==",
             "dev": true,
             "dependencies": {
-                "@types/yauzl": "^2.9.1",
                 "debug": "^4.1.1",
                 "get-stream": "^5.1.0",
                 "yauzl": "^2.10.0"
@@ -8057,7 +8054,6 @@
                 "@types/node": "*",
                 "anymatch": "^3.0.3",
                 "fb-watchman": "^2.0.0",
-                "fsevents": "^2.1.2",
                 "graceful-fs": "^4.2.4",
                 "jest-regex-util": "^26.0.0",
                 "jest-serializer": "^26.6.2",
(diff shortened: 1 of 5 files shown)


site/index.ts
```diff
@@ -159,10 +161,23 @@ domReady(() => {
...
-        });
+        };
+
+        const editorInstance = new StacksEditor(place, content.value, options);
+
         // set the instance on the window for developers to poke around in
         // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-explicit-any
         (window as any)["editorInstance"] = editorInstance;
+
+        const secondEditorInstance = new StacksEditor(
```
hamvocke, line 172: I'm really cool with having a demo page where there are two editor instances since this is not the first bug around multiple editor instances on the same page. I recommend we add the "two editors on the same page" scenario on another site, though. See the `site/variants` directory? That's where we've got a bunch of other demo pages. Let's add the "two editors" scenario there (can just copy the `empty.html` view as a starting point, that should help reduce a lot of clutter). Also let's make sure to link to the new "two editors" page in the "Settings" menu in the top-right corner :)

hamvocke (changes requested):
Verified that the fix works as intended.

Left two comments inline that should hellp make things a little neater.

I wondered if we could add a bunch of tests here and played around a bit. Looks like we can't add unit tests because the `file` input element is really hard to work with in a unit test - but I've got a playwright test coming up that's verifying the entire "close the image uploader" logic, that should help here. I'll add another PR for that test.

src/shared/prosemirror-plugins/image-upload.ts
```diff
@@ -459,6 +459,7 @@ export class ImageUploader implements PluginView {
 
         this.isVisible = isVisible;
         this.image = state?.file || this.image;
+        this.uploadField.value = null;
```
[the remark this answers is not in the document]
WouterDeKort: I've moved the code to resetUploader as you suggested.

site/index.ts
WouterDeKort: I've added a variant named dualeditors, added it to settings and modified index.ts to check if there is a second editor element on the page to avoid null exceptions.

Is it ok how I've done it or did I miss anything?

site/variants/dualeditor.html
```diff
@@ -0,0 +1,25 @@
...
+    <link rel="stylesheet" href="https://unpkg.com/@stackoverflow/stacks/dist/css/stacks.min.css" />
+    <script src="https://unpkg.com/@stackoverflow/stacks/dist/js/stacks.min.js"></script>
+</head>
+
+<body class="p24">
+    <form class="wmx6 m-auto">
+        <textarea id="content" name="content" class="d-none"></textarea>
+        <h1 class="mt32">Dual Editor Playground</h1>
+        <div class="grid">
+            <div id="example-1" class="mr8"></div>
```
hamvocke, line 19: Nit, completely up to you: I think it looks nicer if the editors are stacked on top of each other:

```suggestion
        <div class="grid fd-column">
            <div id="example-1" class="mb8"></div>
```

hamvocke (approved):

package-lock.json
```diff
@@ -1718,7 +1718,6 @@
                 "jest-resolve": "^26.6.2",
                 "jest-util": "^26.6.2",
                 "jest-worker": "^26.6.2",
-                "node-notifier": "^8.0.0",
```
b-kelly, line 1721: Let's revert the changes to this file. Just curious, what's the output of `node -v` on your machine?

site/index.ts
```diff
@@ -133,7 +135,7 @@ domReady(() => {
 
     // asynchronously load the required bundles
     void import("../src/index").then(function ({ StacksEditor }) {
-        const editorInstance = new StacksEditor(place, content.value, {
+        const options: StacksEditorOptions = {
```
b-kelly, line 138: Can we revert this change as well? That would allow us to remove the import at the top. Normally I wouldn't care, but I'm very cautious about adding additional imports to this file so I can keep the bundle separate to test lazy loading.

site/index.ts
```diff
@@ -159,10 +161,22 @@ domReady(() => {
...
-        });
+        };
+
+        const editorInstance = new StacksEditor(place, content.value, options);
+
         // set the instance on the window for developers to poke around in
         // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-explicit-any
         (window as any)["editorInstance"] = editorInstance;
+
+        if ( place2 != null)
```
b-kelly, line 172: In JS, you always want to use `===` and the `!==` equivalent. In this case, we can simply update to:

```suggestion
        if (place2)
```

site/index.ts
```diff
@@ -159,10 +161,22 @@ domReady(() => {
...
+        {
+            const secondEditorInstance = new StacksEditor(place2, content.value, options);
+
+            // set the instance on the window for developers to poke around in
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-explicit-any
+            (window as any)["secondEditorInstance"] = secondEditorInstance;
```
b-kelly, line 178: This looks fine to me. I've set precedence with the absolute hack above 😂.

No changes necessary.

site/index.ts
```diff
@@ -171,6 +185,13 @@ domReady(() => {
             setDefaultEditor(e.detail.editorType);
         }
     );
+
+    place2?.addEventListener(
```
b-kelly, line 189: This is fine as-is, but you could even just leave it out if you wanted. All this does is set the default editor on page load to the last mode you used, which isn't super important for our dual editing test.

src/shared/prosemirror-plugins/image-upload.ts
```diff
@@ -359,6 +359,8 @@ export class ImageUploader implements PluginView {
     resetUploader(): void {
         this.resetImagePreview();
         this.hideValidationError();
+
+        this.uploadField.value = null;
```
b-kelly, line 363: Nice catch!

No changes needed.

src/shared/prosemirror-plugins/image-upload.ts
```diff
@@ -116,7 +116,7 @@ export class ImageUploader implements PluginView {
         this.uploadField.className = "d-none";
         this.uploadField.accept = "image/*";
         this.uploadField.multiple = false;
-        this.uploadField.id = "fileUpload";
+        this.uploadField.id = "fileUpload" + (Math.random() * 10000).toFixed(0);
```
b-kelly, line 119: This looks good for now. We have a concept of an `internalId` on each Editor, but it isn't exposed anywhere or passed down to any of the plugins. I'll think on it, perhaps extracting the method or even potentially adding a linting rule to disallow setting ids to prevent these types of errors in the future.

No changes necessary.

b-kelly (changes requested):
Overall, this looks awesome! Thank you for taking this on. I only have two changes - one is a suggestion and the other is to revert the `package-lock.json file`. Other than that, lgtm!

package-lock.json
WouterDeKort: node -v returns v15.14.0. What version are you using?

I ran node i to get all dependencies and that resulted in updating package-lock.json. Can you help me understand why that happened and why I should revert?

WouterDeKort: I reverted the change I made to this file

site/index.ts
WouterDeKort: But that would mean duplicating the options lines. Just checking if that's what you want? If it's ok, I'll revert the change.

src/shared/prosemirror-plugins/image-upload.ts
hamvocke: In this case we need the `id` though. The `Browse` link in the file upload panel is set up as a `label` using the `for` attribute which (as far as I know) has to point to an `id` in order to work. The only alternative would be to nest the upload `input` within the `label` element.

site/index.ts
b-kelly: Ahhhh, ok that makes sense. It somehow went over my head that you were using it twice. This is perfectly fine then, sorry for the confusion!

src/shared/prosemirror-plugins/image-upload.ts
b-kelly: Right, to clarify better, the linter wouldn't so much be disallowing setting ids, but rather disallowing setting ids without some sort of validation/sanitization. For instance, I just added the [eslint-plugin-no-unsanitized](https://github.com/mozilla/eslint-plugin-no-unsanitized) which does this exact type of thing for e.g. `innerHTML`.

b-kelly (approved):
Nice work! Thanks for making those changes for me. I'll merge this in ASAP.

package-lock.json
b-kelly: Mine is reporting `15.5.0`, so not much farther off yours. I'll update and see if the changes come through on my end as well.

As for the "why", generally we don't want to touch package-lock unless we're updating/adding dependencies to package.json. As for why yours updated the package-lock, specifically? I don't know. npm install can be a little finicky at times and I'm not an expert on why. It may be as simple as "your lock file is right and the committed version is wrong" or "new version, new lockfile rules". I'll take a look into it.

PR #71, opened 2021-05-13
Count characters in the editor content

aalear:
I'm going back and forth on firing an event vs letting the editor do something with the count itself. 

For now, I think this is a bit better for my purposes because it'll let me provide localization and SO-specific event handling at the point where I'll be consuming this. On the other hand, I can also envision some sort of UI built into the editor in the future that accepts additional configuration (e.g. min and max number of allowed characters) and does something fancy with styling.

There may still some quirks with the actual counting. As far as I can tell, this *should* be counting using Markdown rather than HTML, but more testing is needed. For that matter, automated tests are missing and will be added soon assuming we all agree that this approach makes sense in the first place.

### DIFF
--- a/site/index.html
+++ b/site/index.html
@@ -173,6 +173,7 @@ And finally, Stack Snippets:
         <h1>Stacks Editor</h1>
         <h3>Showcasing our new editing experience for Stack Overflow</h3>
         <div id="example-1"></div>
+        <div id="character-count"></div>
     </div>
 </body>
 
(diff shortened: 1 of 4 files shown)


b-kelly (changes requested):
I'm not a huge fan of this approach. As written, this will serialize the document (via `.content`) on _every transaction_. Every keystroke sends a transaction as well as a _ton_ of other processes internally.

If we decide to do this in the editor itself (as opposed to listening to e.g. an input event outside of the editor), we'd also need to make sure it is as performant as possible for the majority of users. This might mean that it's opt-in, calculated periodically, or otherwise.

Just food for thought. I'll think on this a bit as well.

aalear:
Closing this for now in favor of another approach in SO itself.

Issue #33, opened 2021-03-15
Add smart linking (such as SE network links)


b-kelly:
Reported on [Meta.SE](https://meta.stackexchange.com/a/360053/395497).

**Describe the bug**

> Pasting a network link doesn't automatically make it a link in the preview when in rich mode:
> 
> Toggling Markdown On and Off again does make the link look properly formatted:
> 
> Paste link in rich mode:
> ![](https://i.stack.imgur.com/aAIWR.png)
> 
> Link when switched to Markdown mode:
> ![enter image description here](https://i.stack.imgur.com/HtcL8.png)
> 
> Link when switched back to rich editor from Markdown mode:
> ![enter image description here](https://i.stack.imgur.com/cxLgq.png)

PR #72, opened 2021-05-27
Automatically format URLs as links on paste in Rich Text mode

aalear:
Played with adding smarter link paste support as described in #33. I'm not sure if fetching question titles is also a part of that issue, but that work will follow in a separate PR. If that's not necessary, then I think we could consider #33 fulfilled if/when this is merged.

In this PR:

- Detecting when something that looks like a URL is being pasted in and formatting that as a link.
    - No attempt to detect links within a larger body of pasted text. If the whole thing is a link, cool. If not, ah well.
- Pasting into code blocks is left to the default behaviour (or the code paste handler) to sort out.
- Pasting a URL into inline code will treat the pasted content as inline code within the same block.
    - Note that this deviates from how we treat pasting of anything else into inline code (which breaks up the inline code element), but I feel like making _that_ better in the long term is the way to go here. Happy to write it up as an issue if that'd help, but in the meantime I left a `TODO` comment.
- If any text is selected when a URL is pasted in, that text becomes the link text.

I don't know if there's a way to write some automated tests for this (I didn't see any for the code paste handler plugin either), but I tested manually and I'm pretty happy with where this is at. 🤞


b-kelly:
Playing with the functionality locally, it's looking really sharp! There's a bit of funkiness in some of the "pasting into a non-empty selection" edge cases, but the base functionality works really well. I think we should be able to reuse the existing link insertion command to detect where we can/can't paste links, but _that_ code seems to need a bit of work... I'll see what we can do there when I review this PR's code.

In the meanwhile, after some "fun" exploration, I added support for testing "paste" in our unit tests. See the [rich-text/plugins/code-paste-handler](https://github.com/StackExchange/Stacks-Editor/blob/main/test/rich-text/plugins/code-paste-handler.test.ts) tests for examples. Fun fact - neither jsdom nor playwright seem to handle pasting well, so I had to stub it out a bit.

aalear:
@b-kelly Thanks for setting up the testing infrastructure for this. I added a few tests and, in the process, seem to have uncovered a bug in link validation - see https://github.com/StackExchange/Stacks-Editor/pull/72/commits/3e82bbacfef7d259215dc5139f5cae8702e733d5 for details.

All in all, I think this is ready for review whenever you have time.

src/shared/utils.ts
```diff
@@ -162,8 +162,8 @@ export function startStickyObservers(container: Element): void {
 }
 
 // rudimentary link validation that's roughly in line with what Stack Overflow's backend uses for validation
-const validLinkRegex = /^((https?|ftp):\/\/|\/)[-a-z0-9+&@#/%?=~_|!:,.;()*[\]$]+/;
-const validMailtoRegex = /^mailto:[#-.\w]+@[-a-z0-9]+(\.[-a-z0-9]+)*\.[a-z]+/;
+const validLinkRegex = /^((https?|ftp):\/\/|\/)[-a-z0-9+&@#/%?=~_|!:,.;()*[\]$]+$/;
+const validMailtoRegex = /^mailto:[#-.\w]+@[-a-z0-9]+(\.[-a-z0-9]+)*\.[a-z]+$/;
```
b-kelly, line 166: Excellent catch!

No changes necessary

src/rich-text/plugins/link-paste-handler.ts
```diff
@@ -0,0 +1,78 @@
...
+    const { from, $from, to, empty } = state.selection;
+    const schema = state.schema as Schema;
+    if (!empty) {
+        return state.doc.rangeHasMark(from, to, schema.marks.code);
+    }
+
+    return (
+        schema.marks.code.isInSet(state.storedMarks || $from.marks()) !==
+        undefined
+    );
```
b-kelly, line 16: nit: `isInSet` could also return `null`. We can catch both cases by doing the following:

```suggestion
    return !!schema.marks.code.isInSet(state.storedMarks || $from.marks());
```

b-kelly: Side note: Looks like we have this (almost exact) code in a few places now - I'll extract [this function](https://github.com/StackExchange/Stacks-Editor/blob/main/src/rich-text/commands.ts#L578) into utils and update this code to match later.

No changes necessary.

src/rich-text/plugins/link-paste-handler.ts
```diff
@@ -0,0 +1,78 @@
...
+                // except to avoid overriding that default behavior.
+                return false;
+            }
+
+            const link = event.clipboardData.getData("text/plain");
+            if (!link || !validateLink(link)) {
+                return false;
+            }
+
+            if (isInlineCode(view.state)) {
```
b-kelly, line 35: I _think_ you could use [.validContent](https://prosemirror.net/docs/ref/#model.NodeType.validContent) for checking if a node can contain any other given node (w/ marks). I've yet to try it, but it _should_ work for both this, your `code_block` case above, as well as any other invalid situations that could arise in the future.

test/rich-text/plugins/link-paste-handler.test.ts
```diff
@@ -0,0 +1,145 @@
...
+        "should handle pasting non-URL text (%#)",
+        (text) => {
+            const view = createView(createState("", [linkPasteHandler]));
+
+            dispatchPasteEvent(view.dom, {
+                "text/plain": text,
+            });
+
+            const insertedNode = view.state.doc.nodeAt(
+                view.state.selection.from - 1
```
b-kelly, line 38: optional: This can be a bit finicky, as any changes to how this/other plugins handle cursors/pasting will break these tests. You can use my _really ugly_ (but effective) `expect(...).toMatchNodeTree(...)` method to check the contents of the document ([example](https://github.com/StackExchange/Stacks-Editor/blob/main/test/rich-text/plugins/code-paste-handler.test.ts#L106))

test/rich-text/plugins/link-paste-handler.test.ts
```diff
@@ -0,0 +1,145 @@
...
+            expect(insertedNode.marks[0].attrs.href).toBe(text);
+            expect(insertedNode.marks[0].attrs.markup).toBe("linkify");
+            expect(insertedNode.marks[0].attrs.title).toBeNull();
+        }
+    );
+
+    it.each(URLTestData)(
+        "should use existing selection as link text (%#)",
+        (text) => {
+            let state = createState("<p>my example link</p>", [
```
b-kelly, line 73: nit: I'd like a wider variety of tests for this case, if possible. Things like pasting across newlines, pasting with multiple node types selected (e.g. h1 + p), selections with invalid marks/nodes inside of them, etc.

This is not a blocking issue, but let's add a `//TODO` if you decide not to add the additional tests.

test/rich-text/plugins/link-paste-handler.test.ts
```diff
@@ -0,0 +1,145 @@
...
+        }
+    );
+
+    it.each(URLTestData)(
+        "should gracefully paste into a code block (%#)",
+        (text) => {
+            let state = createState(
+                `<pre>int i = 5;
+i++;
+Console.WriteLine(i);</pre>`,
```
b-kelly, line 126: nit: I think this might need `<code>` for the schema to detect as a `code_block`.

```suggestion
                `<pre><code>int i = 5;
i++;
Console.WriteLine(i);</code></pre>`,
```

optional: When I'm testing behavior against very specific nodes, I tend to add a quick `expect` to make sure the nodes I'm testing against appear in my test.

b-kelly (changes requested):
The code looks fine as-is, but could use a few tweaks here and there. I can go ahead and merge this in as-is if `//TODO` comments are added instead (so I remember to come back and do an adjustment pass).

Excellent catch on the url regex! That's a good example of where existing unit tests fall short. Thanks for picking up this functionality as well!

test/rich-text/plugins/link-paste-handler.test.ts
aalear: My original approach here was to do that, but I ran into issues with checking for marks in expected output. I don't recall the specific errors I was hitting, but I spent a fair amount of time staring at the object printout in the browser and matching it with my test setup... and just couldn't get it right. Since you marked it optional, I'd prefer to leave these as is at least for now.

b-kelly: No prob! I bet I know what you ran into too. The way the matcher is currently written, if you include the value, the entire value needs to match. In the case of arrays, this includes all items and their values. Doing something [like this](https://github.com/StackExchange/Stacks-Editor/blob/main/test/shared/markdown-parser.test.ts#L39) should work. That being said, it is ok to leave this as-is.

test/rich-text/plugins/link-paste-handler.test.ts
aalear: Doesn't look like `<code>` is required. I actually had an assert upfront to verify node type, but removed it before submitting. Putting it back in in https://github.com/StackExchange/Stacks-Editor/pull/72/commits/ffd77b340444d9a02c9b214c14f5b6fa8c25d968.

test/rich-text/plugins/link-paste-handler.test.ts
aalear: Ooh yeah, that looks promising. In my attempts, I thought I had the entire tree written out, but I suspect what the browser was showing me was trimmed down.

PR #74, opened 2021-06-08
Switch away from deprecated grid classes

aaronshekey:
`fl1` is getting replaced by the more explicit `fl-grow1` :v:


b-kelly (approved):
lgtm, thanks for the fixes

PR #84, opened 2021-09-21
strip carriage return when pasting into codeblocks

BrianHung:


b-kelly (changes requested):
Thank you for the PR! Generally speaking, I'm ok with the changes presented here, but I would need tests added for the new functionality in order to consider it for merge.

In the meanwhile, I'm going to take a peek at the Prosemirror code you linked and see if there's any thing else in there that we should probably be doing (or better - if we can reuse their functionality a little better).

PR #85, opened 2021-09-27
Do not autolink emails without the "mailto:" prefix

double-beep:
See: [Emails are autolinked in the new Stacks editor](https://meta.stackexchange.com/q/370235) on MSE.


b-kelly (approved):
Nice catch! Thank you for the added test as well

PR #89, opened 2021-10-26
feat(options): add option for default validate link to use markdown-it validate link

bharney0:
Added property to `StacksEditorOptions.parserFeatures` option for allowing users to use the default validate link function from Markdown-It to be used instead of the Stack Overflow's validate link method.

- The default Markdown-It option supports data:image along with a few other link types
image
- Stack Overflow's validate link function only supports http links.
image

While implementing image uploading functionality for our own project, the usage of data:image URLs are not supported by the Stack Overflow validateLink.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-    "name": "@stackoverflow/stacks-editor",
+    "name": "@microsoft/stacks-editor",
     "version": "0.4.1",
     "description": "",
     "repository": {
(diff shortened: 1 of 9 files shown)


bharney0:
Closing this PR out and re-opening without the additional commits. This PR is just to add a defaultValidateLink option to the Stacks-Editor. New PR can be seen here: #91

PR #91, opened 2021-11-01
feat(options): add property to options.parserFeatures to handle custom link validation

bharney0:
Added property to `StacksEditorOptions.parserFeatures` option for allowing users to use the default validate link function from Markdown-It to be used instead of the Stack Overflow's validate link method.

- The default Markdown-It option supports data:image along with a few other link types
image
- Stack Overflow's validate link function only supports http links.
image

While implementing image uploading functionality for our own project, the usage of data:image URLs are not supported by the Stack Overflow validateLink.


b-kelly (changes requested):
Thank you for the feature PR! I'm not opposed to this feature, but I have a few requests before we can merge it.

1. Can you add new tests for this new functionality?
2. `validateLink` is used in a number of places. The change here only checks for validity in the commonmark > rich text conversion. In order to ensure editor-wide validity, we'd also need update the logic in:
   * `inputrules.ts` (translates typed markdown to nodes in rich-text mode)
   * `link-paste-handler.ts` (detects links pasted in rich-text mode and "linkifys" them)
   * `link-tooltop.ts` (used when creating/editing links in rich-text mode)
3. Something to consider would be allowing a user to pass in their own custom validation function. This isn't necessary for merge, but would allow us to plan ahead on the API and not introduce breaking changes later in the project's lifetime.

bharney0:
@b-kelly I added the option to the different areas that are impacted by validateLink

 - `inputrules.ts` added check if option.validateLink exists, then use that instead
 - `link-paste-handler.ts` similar logic to use the option.validateLink if it has been passed in
 - `link-tooltip.ts` now using option if available.

I also added unit tests, but in my opinion they are pretty weak and just test the functionality that my own custom validateLink function returns the result I expected. Let me know what you think.

b-kelly:
@bharney0 Thank you for the changes! I'll take a look at updates this week, likely today or tomorrow.

lamebrain:
@b-kelly Hi Ben! Have you had a chance to review this yet?

b-kelly:
I've decided to go a slightly different direction with this. Implemented in 124762a6. Just need to pass a custom `parserFeatures.validateLink` option.

PR #93, opened 2021-11-23
chore: migrate tests to @playwright/test test-runner

mxschmitt:
Motivation: @playwright/test gives a lot of goodies out of the box: web-first-assertions, projects, better parallelisation, TS out of the box, artifacts handling (tracing, screenshots, videos) which won't get added to jest-playwright-preset in the near future. 

- a lot of the existing assertions can be probably rewritten using [web-first assertions](https://playwright.dev/docs/test-assertions) from Playwright. These include auto retries etc. to make tests less flaky and more reliable.
- The CI pipeline was split up into 3 jobs now, since e2e tests are probably taking longer to run because they are now run on Firefox and WebKit.
- Linting rules with Jest might clash with the way playwright-test does stuff, so it might make sense to either disable them per `*e2e.test.ts` files or move the e2e tests into its own separate folder
- Also traces were enabled which get uploaded as test-artifacts. So if there is a flaky test, they can get downloaded and opened on [trace.playwright.dev](trace.playwright.dev), see [here](https://playwright.dev/docs/trace-viewer).


b-kelly:
Ah, this is really cool! I merged in main to get clear up the merge conflicts and so I could test locally. It works well on my machine. I want to take a second look at the added dependency / config / ci jobs, as well as check the linting rules as you suggest, but otherwise this is looking good to go in.

config/playwright.config.ts
```diff
@@ -1,43 +1,44 @@
...
+            name: "chromium",
+            use: {
+                ...devices["Desktop Chrome"],
+            },
+        },
+        {
+            name: "firefox",
+            use: {
+                ...devices["Desktop Firefox"],
+                // TODO: remove after https://github.com/microsoft/playwright/pull/10492 was released
```
mxschmitt, line 31: we can remove this now and upgrade @playwright/test to version 1.18.0.

b-kelly (approved):
lgtm, this looks great! Added dependencies + configs look good and I modified eslint to only use _some_ of the jest warnings for playwright, since eslint-plugin-playwright doesn't have nearly the same amount of features. GH Actions doesn't take any longer to run, so we're fine there as well.

Thanks for the feature submission!

PR #99, opened 2022-01-24
docs: add disabled view

KyleMit:
Adding more and more menu options doesn't scale super well, but for now, throwing up an example that'll be up at (after merging)

https://editor.stackoverflow.design/disabled.html

Helps us highlight how the disabled editor works.  In a future ticket / PR - we may want to turn off the markdown toggle on a disabled editor - but maybe not too?


aaronshekey (approved):
Looks good to me!

b-kelly (changes requested):
The quick settings menu already contains a disable/readonly toggle. What's the value add here?

image

KyleMit:
Eh - Sorry @b-kelly  - yep, ReadOnly toggle is 100% what I was looking for - closing as dupe

PR #104, opened 2022-03-22
feat: add language dropdown to code_block nodes

### LINKED ISSUE #28: Add the ability to change the code block language in rich text mode
Reported on [Meta.SE](https://meta.stackexchange.com/a/360147/395497).

> This was mentioned in another answer I believe, but I would like to add some mock UX for an idea of how it can work and look.
> 
> *I just added an arrow next to the language to make it look like a selector.*
> 
> 
> When clicking on the language, a dialogue would appear with possible selection. For example, a list similar to the following.
> 
> ```
> ✓ auto
> -------
> c#
> css
> html
> javascript
> plaintext
> python
> ```
> 
> ---
> 
> Another idea would be to show a dialogue similar to the image's details dialogue. Although this can be annoying and would much prefer the option above.
> 
> > enter image description here

b-kelly:
Adds the ability to choose the language for codeblocks in richtext mode. Opening as a draft because all the code is there, but it needs a design pass before being merged.

image

Fixes #28


src/rich-text/node-views/code-block.ts
```diff
@@ -26,7 +26,7 @@ export class CodeBlockView implements NodeView {
 
         this.dom.innerHTML = escapeHTML`
 <pre class="s-code-block"><code class="content-dom"></code></pre>
-<select class="ps-absolute bg-transparent h:bg-unset c-pointer t2 r4 fs-fine us-none js-lang-select"></select>
+<div class="s-select s-select__sm ps-absolute t6 r6"><select class="js-lang-select"></select></div>
```
aalear, line 29: You probably don't want that `</select>` in there now :)

b-kelly: Not sure I follow? `select` is [not a valid self-closing tag](https://html.spec.whatwg.org/multipage/form-elements.html#the-select-element:concept-element-tag-omission), so the end tag is necessary.

b-kelly:
@aaronshekey One of the reasons I didn't go with the `s-select` originally is because it covers too much text. What should we do in cases like this:

image

If possible, I don't want to break the flow of the doc by placing the select outside of the code_block. Perhaps, we could hide that select if the user's cursor isn't in the node or take a similar approach as Jira:

image

src/rich-text/node-views/code-block.ts
aalear: Oh, I misread that whole line. I thought `<select>` was replaced with a `<div>` altogether for some reason. My bad. 🤦‍♂️

b-kelly: Nah, you're good. This is example number 10,134 on "why building html in strings sucks". I have a micro JSX library that is _this close_ to being made public just so I can pull it in to use in this project...

phoebe-leung:
In reference to the JIRA issue and other chat threads in Slack:

@b-kelly I was thinking something like this

image

Much of this is taken from what [Collectives is doing for their tag dropdown](https://stackoverflow.com/collectives/audiobubble?tab=questions) like we talked about in chat. For the code selector though, we’d need to have the dropdown show at the bottom of the code block so that it doesn't block the middle of where people are typing any time they click into a part of the code block. 

In the screenshot above I have it default to no language selected, but it seems like maybe we should have PlainText as the default?

alexwarren:
Superceded by #429

PR #106, opened 2022-03-24
Accessibility improvements for the image uploader

hamvocke:
This PR fixes two accessibility issues on the image uploader:

1. It adds an `alt` text to the preview image after you've selected an image to upload
2. It makes the "browse" button accessible via keyboard (tab navigation), allows triggering the file chooser via keyboard (Enter press on the "browse" button) and automatically focuses the "browse" button once the uploader opens

I'm still unhappy about the outline style used for the browse button. We're using `s-link` under the hood and this is the default style but to me it looks a little out of place. Any ideas?

Screenshot 2022-03-24 at 14 21 51

### DIFF
--- a/src/shared/prosemirror-plugins/image-upload.ts
+++ b/src/shared/prosemirror-plugins/image-upload.ts
@@ -120,7 +120,7 @@ export class ImageUploader implements PluginView {
 
         this.uploadContainer.innerHTML = escapeHTML`
             <div class="fs-body2 p12 pb0">
-                <label for="${this.uploadField.id}">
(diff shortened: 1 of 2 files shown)


b-kelly (approved):
Good looking out! I tweaked your approach a bit such that the actual hidden input itself gets the keyboard focus, so we don't have to try and trigger it in JS (first thing I intuitively tried to do was trigger with SPACE 😛). The `f:outline-ring` class uses `:focus-within` under the hood, so the browse "link" gets a focus outline when the input inside is focused.

PR #109, opened 2022-04-06
Defer triple-click handling to the browser in markdown mode

### LINKED ISSUE #27: Triple-click selects the entire text in Markdown mode
Reported on [Meta.SE](https://meta.stackexchange.com/a/360115/395497).

**Describe the bug**

> It's a standard practice in modern text fields to double-click in order to select a single word and triple-click to select a single **paragraph**. This has worked in the old editor, works in the Rich Text editor, but does not work in Markdown mode and the entire text gets selected instead.
> 
> Selecting paragraph in Rich Text vs Markdown mode

**Expected behavior**

The paragraph the cursor is currently in is selected, not the entirety of the text.

double-beep:
Adds a `Plugin` that prevents prosemirror from applying its triple-click handler function in the markdown mode.

Closes #27
cc @LunarWatcher

### DIFF
--- a/src/commonmark/editor.ts
+++ b/src/commonmark/editor.ts
@@ -60,12 +60,17 @@ export class CommonmarkEditor extends BaseView {
                             props: {
                                 handleDOMEvents: {
                                     mousedown(view, event) {
-                                        const { selection: { $from, $to } } = view.state;
-                                        return $from.sameParent($to) && event.detail === 3;
-                                    }
-                                }
-                            }
-                        })
+                                        const {
+                                            selection: { $from, $to },
+                                        } = view.state;
+                                        return (
+                                            $from.sameParent($to) &&
+                                            event.detail === 3
+                                        );
+                                    },
+                                },
+                            },
+                        }),
                     ],
                 }),
             }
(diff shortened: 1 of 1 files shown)


LunarWatcher:
It's better at least; triple click seems to work like it's meant to. Triple-clicking whitespace still acts a bit weird, but it seems like double-clicking (trailing*) whitespace selects all too :roll_eyes: Double-clicking words still acts fine, not sure if it's worth dealing with

b-kelly (changes requested):
I don't think this is a proper fix to the root issue, but rather fixing one of the symptoms. I've tackled the root cause here a few times, but have yet to implement a proper fix. I'm going to sit on this one a bit as fixing the root cause of this issue is high up on my list.

Either way, this PR needs a few changes before it could be considered for a merge:
* tests for added/changed functionality
* moving the added plugin to its own file
* JSDocs for all added classes, functions, exports, etc

double-beep:
@b-kelly is it OK if I move the plugin into a new `plugin.ts` file (in `src/commonmark`)?

b-kelly:
@double-beep Yep, that sounds fine to me. Usually I'd want to mimic the existing structure in the `rich-text` folder, but I don't see the use in creating a new folder for a single file with a single export. Let's name the file `plugins.ts` and call it good on that front.

double-beep:
How can I add tests for triple clicks? The behaviour seems inconsistent between browsers.

PR #110, opened 2022-04-11
Cycle through headings on Ctrl+H

### LINKED ISSUE #45: Toggle heading shortcut broken in rich text mode
Reported on [Meta.SE](https://meta.stackexchange.com/a/360119/395497)

**Describe the bug**

> ### Pressing <kbd>Ctrl</kbd>+<kbd>H</kbd> does not *toggle* Heading in Rich Text mode
> 
> - In the old editor, it would _cycle_ between several Headings.
> - In the Markdown mode of the new editor, it _toggles_ between Heading1 and no heading.
> - However, in the Rich Text editor, it only works one way. Pressing <kbd>Ctrl</kbd>+<kbd>H</kbd> for a second time opens the browser's history tab in Chrome.

### LINKED ISSUE #105: Heading selection UX improvements
So glad to see headings landed! There's one slightly weird thing about the UX here.

Let's say you've got a Heading 2:

image

When I click Heading 1, it toggles all headings off instead of switching the heading 2 to a heading 1. That interaction should select the Heading 1, not toggle off a heading 2. :v:

🤔

double-beep:
- Toggle code blocks and headings when <kbd>Ctrl</kbd> + <kbd>M</kbd> and <kbd>Ctrl</kbd> + <kbd>H</kbd> are pressed respectively.
- Cycle through headings on <kbd>Ctrl</kbd> + <kbd>H</kbd>.

Closes #45, closes #105


b-kelly (changes requested):
Thank you for the PR! I'll give this a lookover soon, but in the meanwhile, could you add some unit tests in as well? I believe we have some existing command unit tests you could use as inspiration. If not, I can add a few for you to base yours off of.

double-beep:
@b-kelly I have already added an e2e test for Ctrl + H - do I need to add more tests?

src/rich-text/commands/index.ts
```diff
@@ -96,6 +96,51 @@ function toggleBlockType(
     };
 }
 
+const headingInfo = {
+    level: 0,
+    time: 0,
+};
+
+export function cycleThroughHeadings(nodeType: NodeType) {
```
b-kelly, line 104: nit: Add `/** */` docs to all new + exported functions

src/rich-text/commands/index.ts
```diff
@@ -96,6 +96,51 @@ function toggleBlockType(
...
+    return (state: EditorState, dispatch: (tr: Transaction) => void) => {
+        const delay = Date.now() - headingInfo.time;
```
b-kelly, line 106: I'd prefer to not check for delays in our underlying code. I think this is to support successive presses causing cycling, but single presses causing toggle? If that's something we'd like to pursue, let's move that logic into the keybind file instead

b-kelly (changes requested):
Left a few comments on the approach. I'd also like to add this functionality to markdown mode as well.

Regarding unit vs e2e tests - we'd like to prefer unit tests whenever possible, especially for functionality that doesn't need a view to run (as is the case here with the state commands). Unit tests are much faster and can be made much more complex / complete.

PR #116, opened 2022-04-21
Add placeholder to empty input 

dancormier:
### TODO
- [ ] Add placeholder to markdown mode
- [ ] Write a test or two
- [ ] Hide placeholder when is showing "Uploading image…" placeholder
- [ ] Should I use "state" instead of "view"?

### DIFF
--- a/src/commonmark/editor.ts
+++ b/src/commonmark/editor.ts
@@ -71,6 +71,7 @@ export class CommonmarkEditor extends BaseView {
             editorHelpLink: null,
             menuParentContainer: null,
             parserFeatures: defaultParserFeatures,
+            placeholderText: null,
             imageUpload: {
                 handler: defaultImageUploadHandler,
             },
--- a/src/rich-text/plugins/placeholder.ts
+++ b/src/rich-text/plugins/placeholder.ts
@@ -1,0 +1,22 @@
+import { Plugin } from "prosemirror-state";
+import { EditorView } from "prosemirror-view";
+
+// TODO cleanup to match common plugin code style
+export const placeholderPlugin = (text = ""): Plugin => {
+    const update = (view: EditorView) => {
+        // TODO add empty check that accounts for added blocks like blockquote
+        if (view.state.doc.textContent || text === "") {
+            view.dom.removeAttribute("data-placeholder");
+        } else {
+            view.dom.setAttribute("data-placeholder", text);
+        }
+    };
+
+    return new Plugin({
+        view(view) {
+            update(view);
+
+            return { update };
+        },
+    });
+};
--- a/src/rich-text/plugins/placeholder.ts
+++ b/src/rich-text/plugins/placeholder.ts
@@ -6,13 +6,13 @@ import { EditorView } from "prosemirror-view";
 export const placeholderPlugin = (placeholderText = ""): Plugin => {
     const update = (view: EditorView) => {
         const { content, textContent } = view.state.doc;
+        const firstChildType = content.firstChild.type.name;
         // TODO investigate a more dependable check for empty content
         // TODO check for image upload placeholder
         const showPlaceholder =
             placeholderText &&
             !textContent &&
-            (content.firstChild.type.name === "paragraph" ||
-                content.firstChild.type.name === "heading") &&
+            (firstChildType === "paragraph" || firstChildType === "heading") &&
             content.firstChild.childCount === 0;
         if (showPlaceholder) {
             view.dom.setAttribute("data-placeholder", placeholderText);
--- a/src/rich-text/plugins/placeholder.ts
+++ b/src/rich-text/plugins/placeholder.ts
@@ -1,14 +1,23 @@
 import { Plugin } from "prosemirror-state";
 import { EditorView } from "prosemirror-view";
 
+// TODO write a test for this
 // TODO cleanup to match common plugin code style
-export const placeholderPlugin = (text = ""): Plugin => {
+export const placeholderPlugin = (placeholderText = ""): Plugin => {
     const update = (view: EditorView) => {
-        // TODO add empty check that accounts for added blocks like blockquote
-        if (view.state.doc.textContent || text === "") {
-            view.dom.removeAttribute("data-placeholder");
+        const { content, textContent } = view.state.doc;
+        // TODO investigate a more dependable check for empty content
+        // TODO check for image upload placeholder
+        const showPlaceholder =
+            placeholderText &&
+            !textContent &&
+            (content.firstChild.type.name === "paragraph" ||
+                content.firstChild.type.name === "heading") &&
+            content.firstChild.childCount === 0;
+        if (showPlaceholder) {
+            view.dom.setAttribute("data-placeholder", placeholderText);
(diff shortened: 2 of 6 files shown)


src/commonmark/editor.ts
```diff
@@ -8,6 +8,7 @@ import {
     commonmarkImageUpload,
     defaultImageUploadHandler,
 } from "../shared/prosemirror-plugins/image-upload";
+import { placeholderPlugin } from "../rich-text/plugins/placeholder";
```
b-kelly, line 11: Since this is used in both editors, we should move this file to `/shared/prosemirror-plugins`

src/rich-text/plugins/placeholder.ts
```diff
@@ -0,0 +1,40 @@
+import { Fragment } from "prosemirror-model";
+import { Plugin } from "prosemirror-state";
+import { EditorView } from "prosemirror-view";
+
+function showPlaceholder(content: Fragment<any>, textContent: string) {
```
b-kelly, line 5: If you're using vscode, make sure you install the recommended extensions. The project is already set up in such a way that they automatically appear under the Extensions tab as "recommended". In this case, your Eslint extension would have caught the warning the GH action is throwing.

b-kelly: I think we can drop the `any` as the default types already imply it. I'd love to make our ProseMirror types a little more specific, but it's a bit of a pain to do so cleanly.

```suggestion
function showPlaceholder(content: Fragment, textContent: string) {
```

src/rich-text/plugins/placeholder.ts
```diff
@@ -0,0 +1,40 @@
...
+    const { firstChild } = content;
+    const { name } = firstChild.type;
+    const allowPlaceholder =
```
b-kelly, line 8: Hmm... I see where you're going with this, but I'm not entirely sure this is the right approach, since allow-lists can get arbitrarily long. I think we can settle for a slightly worse approach: "Allow any block, so long as the document contains no text". Technically, this will show the placeholder even when the empty document has a styled node opened (such as for a blockquote), but I'm not so sure that's a big deal.

I'll play with it a bit locally and see how I feel about it.

src/rich-text/plugins/placeholder.ts
```diff
@@ -0,0 +1,40 @@
...
+// TODO write a test for this
+/** Plugin that add placeholder text to the editor when it's empty */
+export function placeholderPlugin(placeholder = ""): Plugin {
+    const update = (view: EditorView) => {
+        const { content, textContent } = view.state.doc;
+
+        if (placeholder && showPlaceholder(content, textContent)) {
+            view.dom.setAttribute("data-placeholder", placeholder);
+            if (content.firstChild.attrs.params === "markdown") {
+                view.dom.classList.add("js-markdown");
```
b-kelly, line 25: What exactly is this class for? It looks like we're not targeting it at all and the name isn't very specific.

src/rich-text/plugins/placeholder.ts
```diff
@@ -0,0 +1,40 @@
...
+        name === "paragraph" ||
+        name === "heading" ||
+        firstChild.attrs.params === "markdown";
+    // TODO check for image upload placeholder
+    return !textContent && allowPlaceholder && firstChild.childCount === 0;
+}
+
+// TODO write a test for this
+/** Plugin that add placeholder text to the editor when it's empty */
+export function placeholderPlugin(placeholder = ""): Plugin {
```
b-kelly, line 18: I'd not make this param optional:

```suggestion
export function placeholderPlugin(placeholder: string): Plugin {
```

src/styles/custom-components.less
```diff
@@ -192,6 +192,18 @@
         font-variant-ligatures: none;
     }
 
+    &[data-placeholder]::before {
+        color: var(--black-200);
+        position: absolute;
+        content: attr(data-placeholder);
+        pointer-events: none;
+    }
+    &.js-markdown[data-placeholder]::before {
```
b-kelly, line 201: Don't add styling to `js-*` classes. If you want to target the markdown editor specifically, we can add a base class to that editor to accommodate.

src/rich-text/plugins/placeholder.ts
dancormier: > "Allow any block, so long as the document contains no text".

This was my initial approach, but starting off with blockquote (as you mentioned), ordered lists, and unordered lists looks rough.

<details>
<summary>Examples of blockquote, unordered list with placeholder</summary>

blockquote with placeholder

unordered list with placeholder

</details>

Though not ideal, targeting paragraph/heading/markdown only seems specific enough to be safe IMO but I'm open to other approaches.

src/rich-text/plugins/placeholder.ts
dancormier: I agree that `.js-markdown` could for sure go for a rename (🤦 on `js-*` for this).
This class is to apply some rich text vs markdown styling to the placeholder. Because my approach uses a pseudo-element for the placeholder, I can't style it purely with atomic classes.

src/styles/custom-components.less
dancormier: I removed the JS that sets this `.js-markdown` class and added `.md-editor` as a base class on the commonmark editor in https://github.com/StackExchange/Stacks-Editor/pull/116/commits/8d27d576305e8a193a5733fcc39f25f7e653c2ef. `.md-editor` seems more descriptive/appropriate but I'm still not 100% on it, so suggestions are welcome.

src/shared/prosemirror-plugins/placeholder.ts
```diff
@@ -0,0 +1,51 @@
...
+import { Plugin } from "prosemirror-state";
+import { Decoration, DecorationSet } from "prosemirror-view";
+
+function showPlaceholder(content: Node) {
+    const { firstChild } = content;
+    const { name } = firstChild.type;
+    const allowPlaceholder =
+        name === "paragraph" ||
+        name === "heading" ||
+        firstChild.attrs.params === "markdown";
```
b-kelly, line 11: This causes the placeholder to show in rich-text mode when you load the following markdown:

~~~markdown
```markdown
```
~~~

b-kelly (changes requested):
I made a few tweaks so that the `[data-placeholder]` attribute is added to the first node, rather than the editor root. That way, the text inherits all the font styling from the node without having to add additional classes. Additionally, I added the [aria-placeholder](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-placeholder) attribute to the editor root (where the `contenteditable` attr is).

I'm still not entirely enamored with the allow-list approach. I'll start thinking on better ways to support that.

src/shared/prosemirror-plugins/placeholder.ts
dancormier: Ah, I just realized I never addressed this issue. Looking now.

b-kelly (approved):
lgtm. Made a few slight opinionated tweaks, as well as expanded the scope of the tests a bit

PR #118, opened 2022-04-27
Menu Dropdowns: Assign unique id to prevent clashes with multiple editors

hamvocke:
There's a bug that occurs in combination with menu dropdown items and multiple editor instances.

## The Bug

When I've got multiple editor instances on the same site, menu dropdowns in the second editor instance don't correctly open their corresponding popover. The cause is that Stacks' popover mechanism works based on the `id` attribute and we're not ensuring that dropdown popovers are truly unique ids.

This PR fixes this issue by passing down the editors unique id as part of the editor options so that we can use it to create truly unique ids when generating dropdown menu items.

## Steps to reproduce

1. Go to https://editor.stackoverflow.design/dualeditor.html
2. Click into the second editor instance
3. Click on the "heading" dropdown menu icon
4. See how a popover opens in the first editor instance
5. This issue also existed for our table dropdown menu

Screenshot 2022-04-27 at 09 51 05

## The fix
Before, we generated an `internalId` using `Math.random()` as part of the wrapping editor component and kept it as a `private` field. This `internalId` was mostly used to uniquely identify the "markdown/rich-text" toggle button.

However, we started replicating this `Math.random()` pattern in other places where we needed a similarly unique token to identify DOM elements. We didn't pass down the already generated `internalId` but duplicated the logic, hoping that things wouldn't clash.

With this bugfix, we had yet another place where we need this unique id. Instead of generating yet another random token in place I opted to generate the unique id once and pass it down as part of our `CommonEditorOptions`. The downside is that this will expose this option as part of our public API, on the other hand this might allow us to give users better control for scenarios where, e.g. `labels` that are defined outside of Stacks-Editor should target the `contenteditable` area in our editor.

Here's the fix in action:

Screenshot 2022-04-27 at 09 53 26

You should be able to verify it on the Netlify PR preview.


b-kelly (commented):
I love the fix here, along with the `generateRandomId` function, but I'm not a super big fan of passing the unique editor id around. I'll go ahead and modify this PR to remove that (and merge) to unblock the bug fix. We can revisit in a separate PR if you feel strongly about it.

src/shared/menu.ts
```diff
@@ -278,8 +278,10 @@ export function makeMenuDropdown(
     active?: (state: EditorState) => boolean,
     ...children: MenuCommandEntry[]
 ): MenuCommandEntry {
-    const popoverId = `${key}-popover`;
-    const buttonId = `${key}-btn`;
+    const randomId = generateRandomId();
+    const popoverId = `${key}-popover-${randomId}`;
```
b-kelly, line 282: I made a change here to automatically add an id to all menu dropdowns automatically so consumers don't have to remember to do it themselves.

b-kelly (approved):
lgtm. I removed the global unique id option, added unique ids to menus automatically and found a few id instances you missed. Thanks for spotting and diagnosing this!

PR #121, opened 2022-05-04
fix: treat images as inline

### LINKED ISSUE #53: Images being treated as blocks, but should be treated as inline
Reported on [Meta.SE](https://meta.stackexchange.com/a/360319/395497)

**Describe the bug**

> The following Markdown will produce two images next to each other, but the preview will show them under each other:
> 
> ```
> ```
> 
> 
> This is what the editor shows:
> 

dancormier:
Fixes #53 very simply by rendering images in a `span` instead of a `div`.

I'm fully ready to learn that this is an overly-simple solution but it works fine* AFAICT 🤷‍♂️

---

*It works fine as-in there are no notable regressions. UX is a little lacking in being able move your cursor around freely, but that's a bigger issue and one that I consider out of scope for this  PR.

### DIFF
--- a/src/rich-text/node-views/image.ts
+++ b/src/rich-text/node-views/image.ts
@@ -28,7 +28,7 @@ export class ImageView implements NodeView {
 
         this.popover = this.createPopover();
 
-        this.dom = document.createElement("div");
+        this.dom = document.createElement("span");
         this.dom.appendChild(this.img);
         this.dom.appendChild(this.popover);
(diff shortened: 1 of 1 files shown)


b-kelly (approved):
Nice, that _was_ simple. Good callout on the cursor movement - I bet that's as simple as not forcing input focus when the popover opens

b-kelly:
I opened an issue for the cursor bug: #122

PR #124, opened 2022-05-05
fix: handle hr insert at end of document

### LINKED ISSUE #26: Horizontal rules are auto-selected on insert
Reported on [Meta.SE](https://meta.stackexchange.com/a/360043/395497).

**Describe the bug**

> If you start writing in Rich Text mode and click the ‘Horizontal rule’ button, it’s automatically selected and you can’t deselect it, not even by clicking outside the answer box. Therefore, as soon as you start typing, the horizontal rule disappears.
> 

**Expected behavior**

The horizontal rule should not be selected. A paragraph should also be inserted below it, with the cursor placed in it.

dancormier:
Fixes #26 by detecting if selection is at end of document and inserting a paragraph in addition to the horizontal rule.

This approach feels a little hacky but is the most sensible thing I could come up with. I've created three conditions to properly insert a horizontal rule:

### Empty document

When the document is empty, insert a paragraph and appended with a horizontal rule and a paragraph. This initial paragraph will retain the cursor and the appended paragraph allows for text to be entered after the horizontal rule. 

<details>
<summary>Example gif</summary>

Empty editor hr add

</details>

### End of a populated document

When the selection is at the end of a populated document, insert a paragraph and prepend it with a horizontal rule. This ensures the cursor is placed _after_ the horizontal rule.

<details>
<summary>Example gif</summary>

End of editor hr add

</details>

### Default

This applies to all other cases. This just replaces the current selection with a horizontal rule.

<details>
<summary>Example gif</summary>

Default hr insertion

</details>


src/rich-text/commands/index.ts
```diff
@@ -104,10 +105,38 @@ export function insertHorizontalRuleCommand(
         return false;
     }
 
-    dispatch &&
+    if (!dispatch) return true;
```
b-kelly, line 108: nit: formatting. I'm surprised I don't have eslint or prettier set up to detect this one...

```suggestion
    if (!dispatch) {
        return true;
    }
```

src/rich-text/commands/index.ts
```diff
@@ -104,10 +105,38 @@ export function insertHorizontalRuleCommand(
...
+
+    const docSize = Selection.atEnd(state.tr.doc).$anchor.pos;
+    const { $anchor, $head } = state.tr.selection;
+    const atEnd = docSize === Math.max($anchor.pos, $head.pos);
+
+    if (docSize === 1) {
```
b-kelly, line 114: You may be interested in the [`insertParagraphIfAtDocEnd`](https://github.com/StackExchange/Stacks-Editor/blob/main/src/rich-text/commands/helpers.ts#L11) helper method here 😉

b-kelly (changes requested):

src/rich-text/commands/index.ts
dancormier: Nice! I saw that at one point but completely forgot about it. Thanks for the nudge!

dancormier: This gets rid of the need for that `atEnd` condition but doesn't seem to do the trick for the empty editor condition. I honestly can't say why 🤷‍♂️

double-beep:
I don't think the bug is fixed. If you type `abc` in an empty editor, then insert a horizontal rule, it's still selected. It's not selected if there's already a newline at the end of the document (which is what probably happened in your second gif).

dancormier:
@double-beep nice catch. My initial implementation worked as expected, but I later used `insertParagraphIfAtDocEnd` which introduced a regression. I've removed the usage of `insertParagraphIfAtDocEnd` for inserting a horizontal rule and it now works as expected AFAICT.

src/rich-text/commands/index.ts
dancormier: It could be an issue with my implementation, but using `insertParagraphIfAtDocEnd` reintroduced the issue mentioned in https://github.com/StackExchange/Stacks-Editor/pull/124#issuecomment-1119265518. I've gone back to something close to what I had before using `insertParagraphIfAtDocEnd`.

b-kelly (changes requested):
Let's ensure we've written some unit tests for this. Doing so also solves the "did I cover every use-case" issue 😉.

b-kelly (approved):
lgtm - nice work!

PR #125, opened 2022-05-11
fix: toggle heading only when same level

### LINKED ISSUE #105: Heading selection UX improvements

dancormier:
Fixes #105

Another simple fix that hopefully is not too naïve. Also added a few tests that hopefully give us good heading-related coverage. Once this PR merges, I'd like to add the remaining tests outlined as `it.todo`s.

### TODO

- [x] Add tests

### DIFF
--- a/src/rich-text/commands/index.ts
+++ b/src/rich-text/commands/index.ts
@@ -73,11 +73,11 @@ function toggleWrapIn(nodeType: NodeType) {
 /** Command to set a block type to a paragraph (plain text) */
 const setToTextCommand = setBlockType(schema.nodes.paragraph);
 
-function toggleBlockType(
+export function toggleBlockType(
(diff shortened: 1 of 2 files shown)


double-beep:
Also fixed in #110

dancormier:
@b-kelly seems like linting isn't working as expected on my machine. I'll look into it.

b-kelly (approved):
lgtm! Thanks for the unit tests as well. I made a few tweaks to the wording and whitespace, but this otherwise looks good.

PR #127, opened 2022-05-17
feat: add shortcut labels to tooltips

### LINKED ISSUE #46: Add keyboard shortcuts to tooltips
Reported on [Meta.SE](https://meta.stackexchange.com/a/360152/395497)

> Can we have the toolbar tooltips show the keyboard shortcut for inserting that element into a post, like the previous editor did, so it looks like this:
> 
>

dancormier:
resolves #46

---

After _far_ too much time trying to figure out how to get keyboard shortcuts from [src/rich-text/key-bindings.ts](https://github.com/StackExchange/Stacks-Editor/blob/main/src/rich-text/key-bindings.ts) to correspond to [src/rich-text/commands/index.ts](https://github.com/StackExchange/Stacks-Editor/blob/main/src/rich-text/commands/index.ts), I realized it would probably be fragile at best and impossible at worst.

I think the best way to add these shortcut labels is to just include them on the labels without pulling from anywhere.

### Note

This PR unifies tooltip text mostly by removing prefixed `Insert` (as in `Insert Code Block`).

### TODO

- [x] Add shortcuts to commonmark editor button tooltips
- [x] Figure out shortcut casing
- [x] We can probably remove all the `ctrl + b` and just display as provided (like `Ctrl-B`)

### DIFF
--- a/src/rich-text/commands/index.ts
+++ b/src/rich-text/commands/index.ts
@@ -25,6 +25,7 @@ import {
 } from "../../shared/prosemirror-plugins/image-upload";
 import { richTextSchema as schema } from "../schema";
 import type { CommonViewOptions } from "../../shared/view";
+import { getShortcut } from "../../shared/utils";
 import { LINK_TOOLTIP_KEY } from "../plugins/link-editor";
 import { insertParagraphIfAtDocEnd } from "./helpers";
 import {
@@ -283,31 +284,6 @@ export function exitInclusiveMarkCommand(
     return true;
 }
 
-/**
- * Returns a string containing the label and readable keyboard shortcut for button tooltips
- * @param label The type to toggle to
- * @param mapping? Corresponding command mapping (keyboard shortcut)
- */
-function getTooltipText(label: string, mapping?: string): string {
-    if (!mapping) {
-        return label;
-    }
-
-    let osSpecificMapping = mapping;
-    // Replaces `Mod-` with the OS-appropriate modifier
-    if (mapping.indexOf("Mod-") === 0) {
-        const isMac = /Mac/.test(navigator.userAgent);
-        osSpecificMapping = mapping.replace("Mod-", isMac ? "Cmd-" : "Ctrl-");
-    }
-
-    const readableMapping = osSpecificMapping
-        .split("-")
-        .join(" + ")
-        .toLowerCase();
-
-    return `${label} (${readableMapping})`;
-}
-
 const tableDropdown = () =>
     makeMenuDropdown(
         "Table",
@@ -346,7 +322,7 @@ const tableDropdown = () =>
 const headingDropdown = () =>
     makeMenuDropdown(
         "Header",
-        getTooltipText("Header", "Mod-h"),
+        `Heading (${getShortcut("Mod-h")})`,
         "heading-dropdown",
         () => true,
         nodeTypeActive(schema.nodes.heading),
@@ -383,7 +359,7 @@ export const createMenu = (options: CommonViewOptions): Plugin =>
                 command: toggleMark(schema.marks.strong),
                 dom: makeMenuIcon(
                     "Bold",
-                    getTooltipText("Bold", "Mod-b"),
+                    `Bold (${getShortcut("Mod-b")})`,
                     "bold-btn"
                 ),
                 active: markActive(schema.marks.strong),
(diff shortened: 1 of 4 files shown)


src/rich-text/commands/index.ts
```diff
@@ -462,7 +495,7 @@ export const createMenu = (options: CommonViewOptions): Plugin =>
             {
                 key: "redo",
                 command: redo,
-                dom: makeMenuIcon("Refresh", "Redo", "redo-btn", [
+                dom: makeMenuIcon("Refresh", "Refresh", "redo-btn", [
```
b-kelly, line 498: This one should be left as "Redo"

```suggestion
                dom: makeMenuIcon("Refresh", "Redo", "redo-btn", [
```

b-kelly: Redo and Undo both have keybindings as well, see [here](https://github.com/StackExchange/Stacks-Editor/blob/main/src/rich-text/key-bindings.ts#L48-L50).

src/shared/utils.ts
```diff
@@ -212,6 +212,22 @@ export function escapeHTML(
...
+ * @param mapping? Corresponding command mapping (keyboard shortcut)
+ */
+export function getShortcut(mapping?: string): string {
+    if (mapping.indexOf("Mod-") !== 0) {
+        return mapping;
+    }
+
+    return mapping.replace(
+        "Mod-",
+        /Mac/.test(navigator.userAgent) ? "Cmd-" : "Ctrl-"
```
b-kelly, line 227: Fwiw, here's the regex that `prosemirror-keymap` uses to test for macOS support:

https://github.com/ProseMirror/prosemirror-keymap/blob/master/src/keymap.ts#L5

b-kelly (changes requested):
The approach here looks good. I've thought about this one a lot and I think this is the simplest approach with our current architecture. I have a proof of concept for moving over to a fully plugin based system that (as a side-effect) very closely ties the keybinds to the menu icons, which would negate the soft-linking we're doing here.

Before I can approve for merge however, this PR needs unit tests.

src/rich-text/commands/index.ts
dancormier: Do we ever show Redo and Undo as buttons?

b-kelly: Yup, we show them on small screens + mobile

image

dancormier: Ah of course!

src/shared/utils.ts
```diff
@@ -212,6 +212,22 @@ export function escapeHTML(
...
+ */
+export function getShortcut(mapping: string): string {
+    if (mapping.indexOf("Mod-") !== 0) {
+        return mapping;
+    }
+
+    return mapping.replace(
+        "Mod-",
+        /Mac|iP(hone|[oa]d)/.test(navigator.platform) ? "Cmd-" : "Ctrl-"
+    );
```
b-kelly, line 228: optional: This code is totally, 1000% fine as-is, but I'm feeling pedantic and in the mood for micro-optimizations today:

```suggestion
    if (!mapping.startsWith("Mod-")) {
        return mapping;
    }

    return (/Mac|iP(hone|[oa]d)/.test(navigator.platform) ? "Cmd" : "Ctrl") + mapping.slice(3);
```

test/shared/utils.test.ts
```diff
@@ -142,4 +143,15 @@ describe("utils", () => {
             }
         );
     });
+
+    describe("getShortcut", () => {
+        it("should replace `Mod` with appropriate modifier key for current platform", () => {
+            const shortcut = getShortcut("Mod-z");
+            expect(shortcut).toBe("Ctrl-z");
```
b-kelly, line 150: Will this test fail if run on macOS? 🤔

Can we use Jest to mock `navigator.platform`? If so, we can then test that both macos and "other" scenarios work.

b-kelly (changes requested):

src/shared/utils.ts
```diff
@@ -224,7 +224,7 @@ export function getShortcut(mapping?: string): string {
 
     return mapping.replace(
         "Mod-",
-        /Mac/.test(navigator.userAgent) ? "Cmd-" : "Ctrl-"
+        /Mac|iP(hone|[oa]d)/.test(navigator.platform) ? "Cmd-" : "Ctrl-"
```
dancormier, line 227: I cited "ProseMirror" approach in my commit message. [Here it is](https://github.com/ProseMirror/prosemirror-commands/blob/master/src/commands.ts#L643).

test/shared/utils.test.ts
dancormier: Looks like `navigator.platform` is unset when these tests run. Since the test defaults to applying `Ctrl`, this test was passing. I've added [two tests](https://github.com/StackExchange/Stacks-Editor/pull/127/commits/9cb4159b7f14063c9991fb61ea7a09d58796c38c) that mock `navigator.platform`.

I was tempted to set `SunOS` as the non-Mac platform, but resisted and went with `Win32`.

PR #131, opened 2022-05-20
fix: cycle headings on shortcut

### LINKED ISSUE #45: Toggle heading shortcut broken in rich text mode

dancormier:
fixes #45

---

### TODO

- [x] Tests!


double-beep:
This is so buggy. Toggling off the header is possible only after reaching level 6, then you can't toggle it back on again.

Compare with #110.

dancormier:
> Toggling off the header is possible only after reaching level 6, then you can't toggle it back on again.

Not _exactly_ right. The heading never got toggled off but instead got set to the non-existant `h7`. D'oh! Friday afternoon commits, amiright?

It's been fixed in [7d6d98d](https://github.com/StackExchange/Stacks-Editor/pull/131/commits/7d6d98d4d20143a9df73d63c07c516135e04236d). Thanks @double-beep!

double-beep:
I still believe there should be a delay though. If a header is at level 1 and the user would like to toggle it off, then they'd need to press <kbd>Ctrl</kbd> + <kbd>H</kbd> 6 times, which might be annoying.

src/rich-text/commands/index.ts
```diff
@@ -82,15 +82,23 @@ export function toggleBlockType(
     nodeType: NodeType,
     attrs?: { [key: string]: unknown }
 ) {
-    const nodeCheck = nodeTypeActive(nodeType, attrs);
-    const setBlockTypeCommand = setBlockType(nodeType, attrs);
-
     return (state: EditorState, dispatch: (tr: Transaction) => void) => {
-        // if the node is set, toggle it off
-        if (nodeCheck(state)) {
+        const nodeCheck = nodeTypeActive(nodeType, attrs);
```
b-kelly, line 86: Rather than modify the generic `toggleBlockType` function, let's create a `toggleHeadingLevel` (that ultimately will likely call `toggleBlockType`) and use that instead

b-kelly (commented):
Nice work. I have one pedantic change, but other than that, lgtm. If splitting the header logic out into its own function is a pain, we can merge as-is

b-kelly (approved):
lgtm, thanks for tackling this! I merged in main, fixed the conflicts from my refactoring commit and added an `@internal` flag to the newly exported function (because _some day_ we'll generate type docs for this project...).

PR #132, opened 2022-05-25
ci(config): add PR template

dancormier:
This PR adds a PR template. Simple as that!


b-kelly (approved):
lgtm, thanks for this! I made a few small tweaks to wording and added a checkbox for function docs (as specified in our CONTRIBUTING file).

I'm not exactly sure what the appropriate scope for this would be. For changes to items in the config folder, I used `ci(config): foo`, but use whatever you feel is appropriate here. I'll have to formalize scope naming here before too long to make sure our conventions are consistent.

Issue #21, opened 2021-03-15
Markdown mode highlighting isn't CommonMark compliant


b-kelly:
Umbrella issue to track multiple bugs.

Reported on Meta.SE in the following posts:

* https://meta.stackexchange.com/a/360038/395497
* https://meta.stackexchange.com/a/360050/395497
* https://meta.stackexchange.com/a/360153/395497
* https://meta.stackexchange.com/a/380383/51 - formats based on escaped md symbols
* https://meta.stackexchange.com/a/380381/51 - `lang-*` doesnt work in RT, works on site
* https://meta.stackexchange.com/a/380339/51

Glorfindel83:
Probably related bug when editing https://meta.stackexchange.com/a/363991/295232:
<img width="730" alt="Screenshot 2021-05-03 at 15 04 02" src="https://user-images.githubusercontent.com/24611142/116879785-639c0680-ac21-11eb-801b-061cf83baee5.png">

b-kelly:
@Glorfindel83 Agreed, that looks related to me and is likely due to the mixing of `*` (emphasis) and `**` (strong emphasis). We use highlight.js as the backing syntax highlighter, which uses regex to detect syntax. My guess is that the regex is not perfect in this scenario and is causing this issue.

Without committing to a specific timeline here (stay tuned), this is relatively high on our [priority list](https://github.com/StackExchange/Stacks-Editor/projects/1) to address.

adunndevster:
@b-kelly - I've submitted a fix for this issue into the highlight.js. Maybe you can help promote the PR?

It was SUPER simple. Really the parser was just being over vigilant in trying to match the ending characters for bold/italics.

https://github.com/highlightjs/highlight.js/pull/3456

V2Blast:
I just encountered that bold-and-italics problem myself (well, I noticed it a while ago; I just didn't figure out the source of the problem until now). I also noticed something else: If the bolded-and-italicized text occurs within a blockquote, then the rest of the text from that point onwards is also displayed as if it’s grayed-out (i.e. as if it were part of a blockquote), as well as bolded and italicized.

(Another issue with the Stacks editor’s inline preview/"highlighting" in Markdown mode that doesn't seem to have been mentioned above: It behaves as if the syntax for italics/bold formatting works across multiple lines (e.g. with the opening `*` being on one line and the closing `*` being on a different line) – even though that's not how the post will actually be rendered when the post is saved.

yellis:
[...]

PR #136, opened 2022-06-06
chore: specify highlight.js version

dancormier:
## Describe your changes

This PR updates sets the version to `^11.5.1.`

We were locked at `11.0.1` because it was specified in `package-lock.json` and we using `*` as the version in `package.json`.

## Additional context

This should address some of this issues in https://github.com/StackExchange/Stacks-Editor/issues/21.

### Fixed with this PR

- [Triple asterisks don't terminate](https://github.com/StackExchange/Stacks-Editor/issues/21#issuecomment-831248428)

### Not fixed with this PR

- [Non-terminating italics](https://meta.stackexchange.com/a/360038/395497)
- [Escaped asterisks](https://meta.stackexchange.com/a/360050/395497)
- [Multi-line asterisks](https://meta.stackexchange.com/a/360153/395497)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -40,7 +40,7 @@
                 "eslint-config-prettier": "^8.5.0",
                 "eslint-plugin-jest": "^26.1.5",
                 "eslint-plugin-no-unsanitized": "^4.0.1",
-                "highlight.js": "*",
+                "highlight.js": "^11.5.1",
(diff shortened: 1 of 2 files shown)


double-beep:
ref: https://github.com/highlightjs/highlight.js/issues/3519

PR #140, opened 2022-06-14
fix(markdown-mode): prevent block change from stripping numbers

### LINKED ISSUE #69: Markdown bullet list command strips numbers from the start of lines
Reported on Meta.SE [here](https://meta.stackexchange.com/a/363507/395497).

**Describe the bug**

Using the unordered list command in markdown mode trims numbers from the start of the string.

**To Reproduce**

Add text to **markdown mode** like the following:

```
3 users need to foo the bar
then the bar becomes foo
4 bars have been foo'd
```

Highlight all 3 lines, then click the "bulleted list" command button.

The resulting text looks like:

```
- users need to foo the bar
- then the bar becomes foo
- bars have been foo'd
```

but should look like:

```
- 3 users need to foo the bar
- then the bar becomes foo
- 4 bars have been foo'd
```

dancormier:
fixes #69

**Describe your changes**

This PR prevents markdown mode commands from stripping numbers from the beginning of strings when changing the block type. It achieves this by checking first for ordered list block types then for any non-alpha*numeric* leading characters.

**PR Checklist**

**Environment(s) tested**

- Device: desktop
- OS: Mac OS
- Browser: Firefox 101.0

**Additional context**

The regexes in `matchLeadingBlockCharacters` could for sure use some work. This PR targets the specific issue mentioned in #69 but could be refactored to check leading characters per the markdown spec.

### DIFF
--- a/src/commonmark/commands.ts
+++ b/src/commonmark/commands.ts
@@ -196,7 +196,15 @@ function setBlockType(
 function matchLeadingBlockCharacters(text: string) {
     // TODO this might be too aggressive... remove based on a whitelist instead?
     // TODO HACK assumes all block types are non-letter characters followed by a single space
-    return /^[^a-zA-Z]+\s{1}(?=[a-zA-Z_*[!]|$)/.exec(text)?.[0] || "";
+    // Match ordered lists prefixes
+    let match = /^(\d+)\.\s/.exec(text)?.[0];
+
+    // If text is not an ordered list block, check for other block types
+    if (!match) {
+        match = /^[^a-zA-Z0-9]+\s{1}(?=[a-zA-Z0-9_*[!]|$)+/.exec(text)?.[0];
+    }
+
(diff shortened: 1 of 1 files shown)


b-kelly (approved):
lgtm. Added support for `12)` as well, since `)` is also a valid ordered list marker (see the [spec](https://spec.commonmark.org/0.30/#ordered-list-marker)). I also added the [@internal TSDoc tag](https://tsdoc.org/pages/tags/internal/) as well, which I only recently realized was a thing.

dancormier:
> Added support for `12)` as well, since `)` is also a valid ordered list marker (see the [spec](https://spec.commonmark.org/0.30/#ordered-list-marker)).

In the course of building this, I realized that we should support `)` and neglected to pop a `TODO` in. Thanks for catching this!

It's 🦐-in' time!

PR #144, opened 2022-06-21
fix(commonmark-editor): prevent placeholder clipping

### LINKED ISSUE #143: Markdown placeholder text should wrap
**Describe the bug**
Placeholder text displayed in the markdown editor does not wrap. This results in the placeholder text being clipped when it extends beyond the width of the editor.

<details>
<summary>Screenshots</summary>

### What shows currently

image

### What _should_ show

image

</details>

**Additional context**
I discovered this bug while reviewing https://github.com/StackEng/StackOverflow/pull/12166.

The fix is probably something along the lines of:
```
.ProseMirror pre[data-placeholder]{
    overflow: unset;
    word-wrap: break-word;
}
```

dancormier:
fixes #143 

This is a simple one. Requesting a review just to see if there's any opinions about approach to overflow here (like, we could use `inherit` or `unset` if we wanted to be a little less prescriptive).

Examples with very long placeholder text (`This is placeholder text, so start typing…` repeated like 43 times)
### Before

image

### After
image

### DIFF
--- a/src/styles/custom-components.less
+++ b/src/styles/custom-components.less
@@ -213,7 +213,7 @@
         position: relative;
 
         // ensures markdown placeholder wraps expectedly
-        &.markdown {
+        &.s-code-block.markdown {
             overflow: visible;
         }
 
--- a/src/styles/custom-components.less
+++ b/src/styles/custom-components.less
@@ -212,6 +212,11 @@
     & [data-placeholder] {
         position: relative;
 
+        // ensures markdown placeholder wraps expectedly
+        &.markdown {
+            overflow: visible;
+        }
+
         &::before {
             color: var(--fc-light);
             position: absolute;


src/styles/custom-components.less
```diff
@@ -212,6 +212,11 @@
     & [data-placeholder] {
         position: relative;
 
+        // ensures markdown placeholder wraps expectedly
+        &.markdown {
```
b-kelly, line 216: I'd scope this in a little further even

```suggestion
        &.s-code-block.markdown {
```

b-kelly (approved):
lgtm, thanks for the fix. This fix is fine, but I think we'd be better served by expanding the height of the root codeblock to match the height of the editor frame, but this would require a teensy bit of refactoring that may not be wholly necessary.

PR #146, opened 2022-06-22
feat(commonmark-editor): add preview plugin

### LINKED ISSUE #115: Markdown preview pane support

Some users prefer a read-only preview pane that updates in real-time, side-by-side with the markdown editor. This matches the existing behavior of the legacy Stack Exchange editor.

**Describe alternatives you've considered**

This would be considered an alternative to rich text mode. If the user switches over to rich text mode, the read-only preview should be hidden.

dancormier:
closes #115

This PR adds a preview plugin for markdown mode.

- [x] The preview updates on input change
- [x] Make this opt-in

~~### A thought on this approach~~

~~This PR adds a preview in the form of a read-only rich-text editor. This basically means that there's two editors rendering at once when in markdown mode. It could be worth considering using the existing rich-text editor instance instead and just hiding/showing the markdown editor instead of destroying the editor instance(s) on switching modes.~~


b-kelly:
One feature that I'd like to make sure is added before merge is to have this _off_ by default and is opted into via an editor setting

dancormier:
### A note on the E2E test

The [most recent run](https://github.com/StackExchange/Stacks-Editor/runs/7117786217) shows tests failing in odd ways. I get similar (but a different set) of failures when I run them locally.

src/stacks-editor/editor.ts
```diff
@@ -223,6 +223,16 @@ export class StacksEditor implements View {
         const menuContainerFn = () => menuTarget;
         this.options.menuParentContainer = menuContainerFn;
 
+        // create specific area for the editor menu
+        const previewTarget = document.createElement("div");
+        previewTarget.className = "py16";
+        this.target.appendChild(previewTarget);
+
+        // set the editors' preview containers to be the combo container
```
marradoss, line 231: @b-kelly @dancormier I think this code should be run only optionally, when the preview feature is enabled.

Currently there is a small issue with this div being added and taking some space, even if it's empty:
https://stackoverflow.atlassian.net/browse/PUBPLAT-3427

Alternatively we could adjust the styles of this div. What do you think?

b-kelly: Thanks for the callout. We're currently tweaking how the preview is shown as part of the work for #150, so I'll ensure this is part of the acceptance criteria for that work.

PR #147, opened 2022-06-23
feat(commonmark-editor): detect and format pasted code

### LINKED ISSUE #135: Auto-escape detected code in Commonmark mode

New users that are unfamiliar with Commonmark syntax are pasting code into the editor and becoming frustrated when it does not render as they expected. Additionally, rich-text mode will auto-escape restricted characters when serializing to commonmark. This makes the editing experience more frustrating when users try to fix the post's formatting.

**Describe the solution you'd like**

Text containing code pasted into Commonmark mode should be wrapped in code fences. Text is already parsed to detect code in rich-text mode, so we may be able to extend that functionality to commonmark mode.

The largest issue is that Commonmark mode uses a completely different schema than rich-text mode, which has no idea about node parsing. For every paste, we'd need to spin up a copy of the rich-text schema, parse the paste as nodes, then serialize it back to commonmark. The upside of this is that any pasted html (such as from Google Docs, etc) will get auto-translated to commonmark on paste. The downside is that users may be expecting a plain text paste. The native work around to that problem is for users to use the OS native plain text paste shortcut (ctrl-shift-v on Windows, cmd-shift-v on macOS), which strips out all non-text clipboard entries before pasting.

**Additional context**

User flow issue described on [MSO](https://meta.stackoverflow.com/questions/416801/is-the-new-editors-inline-code-tool-supposed-to-break-up-code-blocks) was used as the basis of this feature request..

dancormier:
closes #135

**Describe your changes**

Detect pasted code and wrap it in code fences.

**PR Checklist**

**Pasting within an existing commonmark code block**

The functionality introduced in this PR does not currently account for pasting into an existing code block in markdown mode (it _always_ wraps pasted code in fences). Although this behavior is somewhat odd from a UX standpoint, it lines up with the existing behavior of the code block command. I'd suggest we create a new issue for this behavior and consider addressing it out of scope for this PR. What do you think @b-kelly?

**A note on the approach taken in this PR**

From the related issue #135:

> For every paste, we'd need to spin up a copy of the rich-text schema[…]

I took a very light touch approach with this PR by purely extending the functionality of `codePasteHandler` just to wrap pasted code in code fences. For the sake of time/scope, I suggest this "light touch" approach for now and revisiting spinning up the rich-text schema/editor as a refactor down the road, since that approach might be applicable in for a few different features (markdown preview pane, rich-text paste node parsing for CommonMark editor, etc.).


src/shared/code-paste-handler-plugin.ts
```diff
@@ -0,0 +1,98 @@
...
+                    return false;
+                }
+
+                // TODO can we do some basic formatting?
+
+                const selectedNode = view.state.selection.$from.node();
+
+                if (isMarkdown) {
+                    codeData = `\n\`\`\`\n${codeData}\n\`\`\``;
+                }
```
dancormier, line 78: This is the meat 🥩

dancormier: Also, those `\n` are not the best approach and I plan to clean this up by detecting if selection is on a new line or not.

src/shared/code-paste-handler-plugin.ts
```diff
@@ -0,0 +1,103 @@
...
+
+                const { $from } = view.state.selection;
+                // TODO can we do some basic formatting?
+                const selectedNode = $from.node();
+
+                if (isMarkdown) {
+                    codeData = `${
+                        $from.pos === 1 ? "" : "\n"
+                    }\`\`\`\n${codeData}\n\`\`\`\n`;
+                }
```
dancormier, line 83: This is the main modification to `codePasteHandler` to operate within the CommonMark editor.

b-kelly:
Made some tweaks to split the plugin into two (one for rich-text, one for commonmark), moving the shared functionality out into a common, testable function. Additionally, I set up the commonmark extension to detect pasted code in the same manner the rich-text editor's schema does.

double-beep:
Since the Commonmark editor is wrapped in a `<code>` block, text that is copied from there and pasted somewhere else is considered code.

b-kelly:
@double-beep Can you be more specific in what issue you're running into? Can you provide repro steps for me?

I've contemplated in the past whether to consider text copied from the commonmark editor as code or not. I'm going to revisit that this morning and make a decision before cutting the next release.

b-kelly:
@double-beep I _think_ I'm reproing the issue you're describing. The fix is simple - I'll get it in place with some updated tests asap.

Issue #152, opened 2022-07-06
Switching between Markdown and Rich Text Editing eats some newlines


tmcentee:
**Describe the bug**
Switching between MD and Rich Text will consistently eat newlines that occur after hardcoded HTML blocks. Subsequent switching between editing modes will continue to eat more newlines further down the page as subsequent text is pulled into the HTML block.

**To Reproduce**
1. Use the following starter text
```
<ul>
<li> Upper block</li>
</ul>

There are spaces above and below this line.

<ul>
<li> Lower Block </li>
</ul>
```

2. Switch between MD and Rich Text mode one time. The first newline above the "There are spaces..." line is now gone.
3. Switch between MD and Rich Text mode one more time. The newline below the "There are spaces..." line is now gone.

**Expected behavior**
Switching between MD and Rich Text should be deterministic.

**Screenshots**
![newline-chomp](https://user-images.githubusercontent.com/17259755/177565976-b21dffe2-8737-404a-83fb-b9617156bbfe.gif)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome
 - Version: 103.0.5060.114 (Official Build) (64-bit)

yellis:
@tmcentee adding https://github.com/StackExchange/Stacks-Editor/issues/152 as a reference here. I am assuming that this is the same issue.

PR #154, opened 2022-07-06
fix(commonmark-editor) ensure newline after html_block

tmcentee:
Addresses part of [152](https://github.com/StackExchange/Stacks-Editor/issues/152). 

**Describe your changes**

This simply ensures two newlines happen at the end of `html_block` nodes. Having other Markdown code on the line immediately after an `html_block` can cause rendering issues when round-tripping between MD and RT editors.

**PR Checklist**

- [X] All new/changed functionality includes unit and (optionally) e2e tests as appropriate
- [X] I've added the `bug`/`enhancement` and other labels as appropriate

**Environment(s) tested**

- Device: Desktop
- OS: Windows 10
- Browser: Chrome
- Version: 103

### DIFF
--- a/src/shared/markdown-serializer.ts
+++ b/src/shared/markdown-serializer.ts
@@ -317,6 +317,8 @@ const customMarkdownSerializerNodes: MarkdownSerializerNodes = {
     // TODO
     html_block(state, node) {
         state.write(node.attrs.content);
+        state.ensureNewLine();
+        state.write("\n");
     },
 
     // TODO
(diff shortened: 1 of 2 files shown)


yellis:
adding https://meta.stackexchange.com/a/380413/51 as a reference here. I am assuming that this is the same issue.

tmcentee:
> adding https://meta.stackexchange.com/a/380413/51 as a reference here. I am assuming that this is the same issue.

That behavior is fixed with these changes.

b-kelly:
Added a more complete fix + tests in 664e034dc39a3ec40fb95e31ba07c925ecbce9cf

PR #156, opened 2022-07-07
chore: fix minor typos

KyleMit:
Just minor spelling fixes - wasn't hurting anybody, but fixing won't either, and then I'll see fewer squiggles in my editor :)

Also adds prettier as the default formatter for all TS files, in-case any users have specified a language-specific formatter of their own

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -16,5 +16,8 @@
     "jestrunner.configPath": "config/jest-unit.config.js",
     "[html]": {
         "editor.defaultFormatter": "esbenp.prettier-vscode"
-    }
+    },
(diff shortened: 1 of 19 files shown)


dancormier (approved):
Tanks! You're langauge substitions really tighted up this codebase since it constained many arbitary tpyos. Much better afer! 😜

--

Seriously though, thanks!

b-kelly:
Thanks for this! You can tell that sometimes I type faster than I think 😅. Do you have a preferred spell-checking extension for VS Code? Currently, mine doesn't do spell-checking, as you can clearly see.

@dancormier That comment hurt my brain. Well done.

PR #157, opened 2022-07-07
fix(rich-text-editor): allow any block inside list item

### LINKED ISSUE #63: Blockquote, etc buttons aren't enabled in list items
**Describe the bug**

When inside a list item, blockquote, code_block, etc toolbar buttons are disabled. However, pressing `enter` to get a new item, then `backspace` to make it a paragraph within the _current_ list item allows you to add block items.

**To Reproduce**

1. Switch to rich text mode
2. Type `* ` to get a new list item
3. Note that blockquotes, etc are not enabled

**Expected behavior**

We should be allowed to insert blockquotes into list items, wrapping in an enclosing paragraph if we have to.

KyleMit:
Closes #63

### Describe your changes

Schemas dictate which elements are allowed to appear under each node.  The previous schema for list item content was `paragraph block*`, meaning the first child must be a paragraph then 0 or more blocks under that.

If we want to allow blockquotes *directly* inside a list item, we'll have to open up and just allow `block+`, meaning one or more of any of the block level elements that we've allowed - including paragraphs, as well as blockquotes, codeblocks, etc.

### PR Checklist

- [x] All new/changed functions have `/** ... */` docs (no changes to top level functions)

### Additional context

Change is a somewhat simple one here.  One thing I wanted to investigate in making this change is how other rich text editors handle this in the UI / Action.

* **Jira** - Will exit list, create new-line, and start blockquote there
* **Reddit** - Will exit list, and convert list item to blockquote
* **Default** ProseMirror Editor - does not allow, because default schema only allows paragraph items as the direct child of a list item

So support for this specific feature elsewhere is not widespread.  We already do allow someone to create a new line within a list item and insert a blockquote on it's own line.

That said, if our goal is to ever have parity between markdown editor and rich text editor, any content restrictions will hurt our ability to do that.  In markdown, you are certainly allowed to type `* > blockquote list item` - so anyone starting from markdown can get themselves into a situation that our Rich Text schema would consider "invalid" if we have strict schema policies.  So it's probably better to relax them, even if it leads to some odd (_but perfectly valid_) markup


src/rich-text/commands/index.ts
```diff
@@ -37,7 +37,7 @@ import { _t } from "../../shared/localization";
 export * from "./tables";
 
 //TODO
-function toggleWrapIn(nodeType: NodeType) {
+export function toggleWrapIn(nodeType: NodeType) {
```
KyleMit, line 40: needed to test blockquote command which is generated via this method.  Other commands (like `toggleHeadingLevel`) are exported from here as well, so no over-arching reason to keep private

b-kelly: If you expose this method, you need to write documentation and unit tests for it 😉

b-kelly (changes requested):
This change seems appropriate to me. You'll need to add new tests and documentation for the newly exported function though.

src/rich-text/commands/index.ts
KyleMit: so, in a sense, this is actually the system under test when testing blockquotes can be inserted into lists.  What we want to check is a) can we execute this command and b) does it apply the right transform.  So there's a single existing test for it, but I added some more test cases for some basic coverage of toggling back and forth

KyleMit:
@b-kelly - should be all set - added node tree string to custom matchers

b-kelly (commented):
I'd like to add some parsing/serialization tests here as well. I'll do this later today before merge. Otherwise, I think this all looks good

src/rich-text/schema.ts
```diff
@@ -169,7 +169,7 @@ const nodes: {
     },
 
     list_item: {
-        content: "paragraph block*",
+        content: "block+",
```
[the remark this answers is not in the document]
b-kelly: Just double checked the commonmark spec for _absolute correctness_™, and it seems that list items allow basically any block element that isn't a horizontal rule, so this is good to go.

https://spec.commonmark.org/0.30/#list-items

KyleMit:
@b-kelly, biggest gap was that there was a `<code>` block, but not the actual text inside it.  Also some white space stuff - valid markup to just adding four spaces to create a code block, even in a nested list - and that's what the markdown serializer is doing, so adding that into the test.  Should be fixed now

b-kelly:
@KyleMit 🤦 Wow, that was a dumb mistake on my part! I _knew_ the test _should_ be working, but I guess I was too braindead at the time to realize it was that simple. I'll give this one last pass today and merge.

PR #158, opened 2022-07-07
feat(rich-text-editor): Add tagLink, subscript, superscript, kbd, and spoiler input 

### LINKED ISSUE #51: Add rich text insert equivalents of all markdown-only nodes
Related report on [Meta.SE](https://meta.stackexchange.com/a/360588/395497)

There is no way to create many custom markdown-only nodes in rich-text mode. These include:
* spoilers
* tag badges
* html-only nodes like `<sub>`, `<sup>`, `<kbd>`

**Describe the solution you'd like**

Add in a way to add these via rich text. Shortcuts are a given, but perhaps we could add some buttons to the UI somewhere that isn't overly intrusive or cluttered.

tmcentee:
Closes #51. 

**Describe your changes**

This PR adds four new features to the Rich Text editor that were previously Markdown-only. All include keyboard shortcuts as well as manual entry via a new overflow dropdown menu option.

These include the following:

## TagLinks <kbd>CTRL-[</kbd>

insert-tag

## Subscript <kbd>Ctrl - ;</kbd> and Superscript <kbd>Ctrl - :</kbd>
subsup

## Spoiler <kbd>CTRL-]</kbd>
spoiler

**PR Checklist**

**Environment(s) tested**

- Device: Desktop
- OS: Windows
- Browser: Chrome
- Version: 103


aalear:
The tag link shouldn't support multiple words. There's existing validation for allowed tags in `parse_tag_link()` that could perhaps be reused?

tmcentee:
@aalear Thanks for the heads-up. I pushed some changes to better validate the links. Now you can only convert the selection if the trimmed content doesn't have any whitespace. (trimmed because double clicking words in Chrome on Windows also selects the space after the word)

test/rich-text/commands/index.test.ts
```diff
@@ -412,6 +534,8 @@ describe("commands", () => {
         it.each([
             [`middle of some text`, false],
             [`<em>cannot exit emphasis from anywhere</em>`, true],
+            [`<sup>cannot exit emphasis from anywhere</sup>`, true],
+            [`<sub>cannot exit emphasis from anywhere</sub>`, true],
```
b-kelly, line 538:
```suggestion
            [`<sup>cannot exit sup from anywhere</sup>`, true],
            [`<sub>cannot exit sub from anywhere</sub>`, true],
```

src/rich-text/commands/index.ts
```diff
@@ -421,6 +421,43 @@ const headingDropdown = (schema: Schema) =>
         )
     );
 
+const overflowDropdown = (schema: Schema, options: CommonViewOptions) =>
```
b-kelly, line 477: nit: needs docs

src/rich-text/commands/index.ts
```diff
@@ -141,6 +145,55 @@ function getHeadingLevel(state: EditorState): number {
...
+        }
+
+        if (!dispatch) {
+            return true;
+        }
+
+        let tr = state.tr;
+        const nodeCheck = nodeTypeActive(state.schema.nodes.tagLink);
+        if (nodeCheck(state)) {
+            const selectedText = state.selection.content().content.firstChild
```
b-kelly, line 176: This fails when the tag is not the first child in the slice, most notably, when a tag falls in the center of other text:

https://user-images.githubusercontent.com/14169651/178818641-1c43eda6-b35f-48c0-8ca3-a5aef92542a7.mp4

src/rich-text/commands/index.ts
```diff
@@ -421,6 +474,43 @@ const headingDropdown = (schema: Schema) =>
         )
     );
 
+const overflowDropdown = (schema: Schema, options: CommonViewOptions) =>
+    makeMenuDropdown(
+        "Overflow",
+        _t("commands.overflow"),
+        "overflow-dropdown",
+        () => true,
+        () => false,
```
b-kelly, line 483: Currently, this causes the dropdown to never be set as active, even when the cursor is otherwise inside an active child.

src/rich-text/commands/index.ts
```diff
@@ -421,6 +474,43 @@ const headingDropdown = (schema: Schema) =>
...
+        dropdownItem(
+            _t("commands.tag", { shortcut: getShortcut("Mod-[") }),
```
b-kelly, line 485: Let's standardize on the term "tagLink" here as "tag" could be otherwise confusing.

src/rich-text/commands/index.ts
```diff
@@ -421,6 +474,43 @@ const headingDropdown = (schema: Schema) =>
...
+            toggleTagLinkCommand(options.parserFeatures.tagLinks.allowNonAscii),
+            "tag-btn",
+            nodeTypeActive(schema.nodes.tagLink),
+            ["fs-body1", "mt8"]
```
b-kelly, line 489: Let's remove all these extra classes - I'd prefer if all of our menus looked the same by default. The reason the heading dropdown has these classes is because it is setting the size to visually simulate the differences in headings. I wouldn't mind setting spoiler, sub, sup classes on these, but that'd also set the styling on the shortcuts as well 🤔

src/rich-text/commands/index.ts
```diff
@@ -421,6 +474,43 @@ const headingDropdown = (schema: Schema) =>
         )
     );
 
+const overflowDropdown = (schema: Schema, options: CommonViewOptions) =>
+    makeMenuDropdown(
+        "Overflow",
```
b-kelly, line 479: "Overflow" is not descriptive of the contents. Maybe "More" or "Misc"? What do other editors have for this text?

test/rich-text/commands/index.test.ts
```diff
@@ -390,6 +392,126 @@ describe("commands", () => {
         });
     });
 
+    describe("toggleTagLinkCommand", () => {
```
b-kelly, line 395: I don't see any tests for un-toggling the taglink or any tests for the parameters you can pass into it. Additionally, we might want to do some active/validity checking when trying to insert taglinks into nodes where they aren't allowed.

b-kelly (changes requested):
Thanks for the PR! I have added some suggestions for changes inline

src/rich-text/commands/index.ts
b-kelly: On second thought, don't worry about this. I think I'll extend the menu helper to do this automatically.

src/rich-text/commands/index.ts
tmcentee: The issue here was trying to validate the selected text as a tagName every time we were toggling, when we should only do so when we're about to create a tagLink. Moved the validation logic down a level and it works properly now.

src/rich-text/commands/index.ts
tmcentee: Jira does "More formatting" so I went with that.

tmcentee:
I pushed updates to the keyboard shortcuts for superscript and subscript to match most other editor conventions.

b-kelly (commented):
Note - blocked by #157, which also exposes `toggleWrapIn`, adding tests and docs

src/shared/utils.ts
```diff
@@ -306,3 +307,24 @@ export function bindLetterKeymap(
         [prefix + letter.toUpperCase()]: command,
     };
 }
+
+/**
+ * Tests whether a string is a valid tag name.
+ * @param tagName The string to test
+ * @param allowNonAscii Whether to allow non-ascii characters in the tag name
+ */
+export function validateTagName(
```
b-kelly, line 316: General tag name validation option was recently added in 6ae693694c08be9afbad7f0668e0043ec747cef2 for a completely unrelated reason, but works great for this use case. Let's drop this method and use that one instead.

b-kelly (changes requested):
Imo, the rich-text version is ready for merge. Any chance we can get the dropdown menu in markdown mode as well? I'd like to keep the menus as aligned as possible. If this is an outsized request, we can consider getting this PR merged first.

tmcentee:
> If this is an outsized request, we can consider getting this PR merged first.

I think we should get this merged and follow up with the Markdown menu items. I don't have a ton of time this sprint to dig into it myself.

PR #162, opened 2022-07-12
fix(commonmark) nested sub tags getting lost when round-tripping

### LINKED ISSUE #32: Fix html rendering issues
Umbrella bug report to track general html parsing and rendering issues.

Reported on Meta.SE in the following posts:

* https://meta.stackexchange.com/a/360181/395497

yellis:
Closes #32 

Thus far:

- Added failing test
- Fixed issue with `sanitizeInlineHtmlTokens` to ensure that it doesnt consider a second open tag to be the closing tag for the first matching open tag

@b-kelly from what I was able to determine, this is still failing however, since when it is converted into the prosemirror `doc`, it becomes a `node` with the text as the content and `sub` as the attribute. And seems that it just cant figure out two open tags of the same type.

Looking for advice on where to go from here - assume that I need to add to the prosemirror the ability to recognize this. Does this mean a schema change, to allow a sub to have a child sub? Need to get it recognized both into PM and out of it back to markdown. If there is prior PRs that do this sort of thing that I can study, would appreciate it. Can talk about this in our scheduled time tomorrow.

### DIFF
--- a/src/rich-text/schema.ts
+++ b/src/rich-text/schema.ts
@@ -517,8 +517,8 @@ const marks: {
         excludes: "",
         attrs: {
             nesting: {
-                default: null
-            }
+                default: null,
+            },
         },
         parseDOM: [{ tag: "sup", getAttrs: nesting("sup") }],
     },
@@ -528,11 +528,11 @@ const marks: {
         excludes: "",
         attrs: {
             nesting: {
-                default: null
-            }
+                default: null,
+            },
         },
         parseDOM: [{ tag: "sub", getAttrs: nesting("sub") }],
-    }
+    },
 };
 
 // for *every* mark, add in support for the `markup` attribute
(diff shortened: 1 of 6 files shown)


b-kelly:
@yellis I've committed a partial fix along with a few other tests I wrote to help me debug the issue. There are really a few problems here:

1. opening a mark inside another mark of the same type does not work unless their attributes are different
   - fixed - I'm keeping track of the nesting and setting that value (might be a 50% fix though if there are more sibling marks of the same type at the same nesting level)
2. `closeMark` is not taking attributes into account when closing a mark
   - upstream bug - the method is using `MarkType.removeFromSet` instead of `Mark.removeFromSet`. The difference being that the former removes all marks of the same type while the latter checks the equality of the Marks before removing them

I'll submit a PR upstream explaining the rationale along with a fix.

yellis:
@b-kelly were you able to submit the upstream PR?

yellis:
@b-kelly were you able to submit the upstream PR?

b-kelly:
Sorry for the delay - this has really gotten away from me. I created an internal Jira ticket (STACKS-121) and tagged you on it so we can assess this during our next backlog grooming session.

yellis:
I am closing this as there is now an internal Jira ticket to track it.

PR #177, opened 2022-07-13
feat: support copy/pasting tag link nodes in rich text mode

### LINKED ISSUE #56: Ensure that custom nodes are serialized/deserialized to/from html properly
Umbrella bug report to track general node serialize/deserialize issues. Serialize is important for copying nodes to the clipboard in a way that we can deserialize on a later paste event. Ideally, custom nodes should serialize as closely to their final rendered versions as possible so that the deserialize can handle both "copied from rich text mode" and "copied from a final rendered post".

Reported on Meta.SE in the following posts:
* https://meta.stackexchange.com/a/360079/395497

aalear:
This PR addresses tag link serialization for #56.

The core of the change is switching from NodeView to a schema definition with appropriate toDOM/parseDOM implementations. 

It looks like we never wrote tests for the original NodeView nor do we have existing tests for other schema specifications. I'm happy to take a stab at adding some, but for now I left them out.

Also note that pasting from a final rendered document is not currently supported. Part of the issue is that public Q&A still uses `.post-tag` rather than `.s-tag` from Stacks. I'm not sure if it's worth adding `.post-tag` checks into the editor repo. Feels like we should be moving towards making the Q&A rendering use Stacks classes instead, although that's also a non-trivial undertaking.


aalear:
@b-kelly I re-tested this the other day and I think everything is good to go. No feature-related changes needed after #158 went in.

b-kelly (changes requested):
TODO needs tests

I'll add these asap so we can merge.

dancormier:
Closing this since it's been > 4yrs since the last update. Feel free to reopen and/or open an issue if it's appropriate

PR #217, opened 2022-08-24
markdown three-way preview toggle

yellis:
Per earlier discussions, introduces 3-way toggle for display mode into the preview option

image

The `StacksEditor:view-change` event on the editor instance now returns an object with both `editorType: number` (unchanged) and `previewShown: boolean`.


b-kelly (changes requested):
TODO needs tests
I'll add these asap so we can merge.

PR #221, opened 2022-09-11
fix(preview-plugin): Reinitialize Markdownit renderer with passed-in options

yellis:
Addresses issue holding up integration of editor preview in main core code.

Issue is that while the `DeepMerge` call that is made at the [top of the main constructor](https://github.com/StackExchange/Stacks-Editor/blob/c8714dfa7912a5c05e45fcb8da638644a07ed07a/src/stacks-editor/editor.ts#L71) for the editor copies all of the properties for the `MarkdownIt` instance being passed in at `options.commonmarkOptions.preview.renderer`, it does not copy in the prototype functions that are part of the object. So when the preview plugin tries to [use the renderer to render text](https://github.com/StackExchange/Stacks-Editor/blob/c8714dfa7912a5c05e45fcb8da638644a07ed07a/src/commonmark/plugins/preview.ts#L108), the `render` function is not present, and an error like this occurs.

image

**Describe your changes**

Changed the initialization of `this.renderer` in the main constructor so that in the case where a `MarkdownIt` renderer is passed in through `options`, it will use the passed-in options to initialize a new `MarkdownIt` object (which will have the relevant prototype functions attached).

Not quite sure how to test this through a unit test.

**PR Checklist**

**Environment(s) tested**

- Tested on chrome latest, Windows 10
- Fix is confirmed to work on local for importing an external `MarkdownIt` renderer and using it in Core code.

### DIFF
--- a/src/commonmark/plugins/preview.ts
+++ b/src/commonmark/plugins/preview.ts
@@ -42,9 +42,11 @@ class PreviewView implements PluginView {
         this.dom = document.createElement("div");
         this.dom.classList.add("s-prose", "py16", "js-md-preview");
         // TODO pass down the ExternalPluginProvider as well
-        this.renderer =
-            previewOptions?.renderer ||
-            createDefaultMarkdownItInstance({
+        this.renderer = previewOptions?.renderer 
+            // if renderer is passed in, need to reinitialize the renderer to ensure that prototype functions 
+            // that were excluded from deepmerge are included in the renderer being used here
+            ? new MarkdownIt(previewOptions.renderer.options)
+            : createDefaultMarkdownItInstance({
                 ...parserFeatures,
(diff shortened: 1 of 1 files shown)


b-kelly:
Good catch - I'll keep this in mind for future options I add. As discussed internally, I'm going to replace this option with a `render` function, so I'm going to hold off on merging/closing this PR until that change is made.

yellis:
@b-kelly should the custom renderer just use the same property? And stacks editor would assume it is a function with signature `(input: string) : string`, that can be called directly as `var html = this.renderer(text)`?

b-kelly:
@yellis I just added support for arbitrary rendering in 13bc4df8b3e9d2781faf801d46511578cb38b5a9. Closing this PR as the change is no longer relevant.

PR #223, opened 2022-09-15
feat(a11y): add additional accessibility attributes on the contenteditable

balpha:
This gives the editor a multiline textbox role, see https://www.w3.org/TR/wai-aria-1.1/#textbox.

It also allows specifying to additional view options (`id` and `ariaLabeledBy`) so that users can programmatically associate the editor box with its label or other elements.

For reference, this is necessary to fix item 726108 in the accessibility audit (WARP-71).

This PR does not include unit tests since the added functionality is just about modifying DOM elements. It does include e2e tests however.

I've added to additional view options, `id` and `ariaLabeledby`, but I'm also open to changing that to something like a generic `attributes` option if you prefer that.


b-kelly:
I liked your suggestion of being able to set arbitrary attributes, so I went ahead and implemented that 
in 8220dd6a3a317578b574a800f8b3dfbb65a0beb4 (to shorten back-and-forth comms here). You can use it via `options.elementAttributes` as follows:

```js
new StacksEditor(..., {
    // ...
    elementAttributes: {
        id: "foo",
        ariaLabeledby: "bar",
    }
})
```

Please let me know if this is not a solution for your intended use case. Closing as implemented.

b-kelly:
@balpha Ah, agreed on those a11y attributes. I forgot to do those in that commit - I'll add them in a separate commit today.

b-kelly:
a11y attributes added in a3631201099c528b3995fed1df34577b96b042fe. Thanks for the reminder!

PR #232, opened 2022-11-02
Add distinct Code of Conduct file

b-kelly:
Having a distinct Code of Conduct file is a qualifier for Netlify's [open source plan](https://www.netlify.com/legal/open-source-policy/). We already lay out our code of conduct in the [CONTRIBUTING](https://github.com/StackExchange/Stacks-Editor/blob/main/CONTRIBUTING.md) file, but it doesn't hurt to have it split out as well.

Looking for a quick double check on this - does the content make sense the way I worded it? Once we're happy with the text, I'll copy this file to the other Stacks repos as well.


b-kelly:
Got buy-in from the CMs that this abridged version is compatible with the broader CoC as a whole. Merging!

PR #235, opened 2022-12-08
fix(code-blocks): ignore single backticks when copy/pasting code blocks

### LINKED ISSUE #230: bug(markdown) Code blocks surrounded by single backtick then triple backticks
**Describe the bug**
Pasting fenced code block into inline code block results in the following being saved
image

A side issue here might be that the inline code block menu button has much more prominence than the fenced code block button, so users go to it first and dont think of fenced code block (could also be something that the interface is expecting too much from the user here).

**To Reproduce**
1. Create inline code block (can use menu)
2. Paste in fenced code block
3. Note: the inline code block ticks are still there, and if saved, will be orphaned ticks surrounding the fenced code block.

**Expected behavior**
Pasting a fenced code block into an inline code block results in only the fenced code block being inserted, and the preceding and following inline code block ticks being removed.

**Screenshots**

Windows 10 Chrome Latest

**Additional context**
[Code blocks surrounded by single backtick then triple backticks](https://meta.stackoverflow.com/questions/421118/code-blocks-surrounded-by-single-backtick-then-triple-backticks). Answer on that page has further repro info.

giamir:
closes https://github.com/StackExchange/Stacks-Editor/issues/230

**Describe your changes**

This PR addresses an issue where users copy/paste "code blocks" (triple backticks delimiter) inside an "inline code block" (single backticks delimiter). For more context about why this user behavior was happening refer to [this meta thread](https://meta.stackoverflow.com/questions/421118/code-blocks-surrounded-by-single-backtick-then-triple-backticks).

The solution solves the issue by checking if the editor view selection is between backticks before appending the clipboard code data. Code feels a bit ad-hoc but I could not find a different way to tackle the issue which would not require large refactors (at least is only 5 extra lines or so). I am open to suggestions.

**PR Checklist**

### DIFF
--- a/src/shared/prosemirror-plugins/code-paste-handler.ts
+++ b/src/shared/prosemirror-plugins/code-paste-handler.ts
@@ -1,4 +1,4 @@
-import { Plugin } from "prosemirror-state";
+import { Plugin, EditorState } from "prosemirror-state";
 import { EditorView } from "prosemirror-view";
 import { Slice, Node, DOMParser, Schema } from "prosemirror-model";
 import { richTextSchemaSpec } from "../../rich-text/schema";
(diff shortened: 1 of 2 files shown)


dancormier (approved):
This PR fixes the _specific_ problem mentioned in the Meta post, but leaves a another issue in place. If a user enters anything into a inline code block, then pastes code, the same issue occurs. 

**Example**:



I'm approving this pull request despite the above issue because this PR addresses the issue mentioned. Also, I anticipate that the issue I found is _way_ less common than the one mentioned in the Meta post. I'll defer to you @giamir to decide if you want to consider the above issue in scope, but I'm happy with this fix and think it would be beneficial to ship.

yellis:
@dancormier @giamir could the logic be extended so at if the _only_ thing in the inline section is white-space, it will still work (and replace the white-space)? Scenario: 

1. inline-code button
2. type a space
3. paste in code

You end up with just the code section?

giamir:
Thanks @dancormier and @yellis for the review. I have now addressed the case when the user is copy/pasting code between backticks containing whitespaces (as requested by @yellis). I don't think we should make any other smart correction since it could be frustrating for the user to lose content when pasting code.

```
` some content  (cursor here)`
``` 
In scenario like the one described above no content will be stripped away when code is pasted (backticks and text will persist).

b-kelly (changes requested):
This only fixes a very small subset of potential code pasting errors. While reviewing this PR my kneejerk reaction was "why should we be in the business of fixing others' bad markup?" and it struck me that you could say the same for auto-fencing pasted code as well.

I'd like to propose that instead of playing whack-a-mole with edge cases here, we simply disable auto-code fencing in markup altogether (but leaving for rich-text as that's so much simpler to support).

I'll add a commit to disable auto-fencing in markdown mode today.

<details><summary>some edge cases</summary>
What happens when we paste in the following situations? Assume the 📋 emoji is where the paste event happens. These are just a few common edge cases that I can think of off the top of my head. I'm sure I could find some more esoteric ones if I revisit the commonmark spec.

~~~md
Inside a code block (also applies to fences using tildes)
```
📋
```

What about this code block with no close? (yes, that's a valid commonmark codeblock)

```
📋
~~~

~~~md
Code spans can be multiple ticks: https://spec.commonmark.org/0.30/#code-spans

`` `📋  ``

Escaped ticks aren't considered either:

\`📋  `code span`
~~~

</details>

b-kelly:
Support removed in 008d0f3796e405ce6d1a065e39146979c8ae5b28

yellis:
> why should we be in the business of fixing others' bad markup?

@b-kelly because in cases where we are sure that this is what they intended, it makes for a better user experience.

Specifically in the case where someone pastes a block of code into an inline code block, that they almost certainly added in because that was the obvious "code" button and they missed the fenced code block button - it is almost certain that they want the fenced code block. And the shortcoming may be in the design of the buttons, or the buttons not being adaptive to context (if you click the "code" button while on a new line, one might expect it to be for a code block, while if you are in a paragraph you would want it to be inline).

While I appreciate the hard stand of "not my fault if users screw things up", the ultimate result of this will be posts with a single tick, followed by lines of unformatted code, followed by a single tick (arguably worse off than it was before this bug was reported, where at least they had the code block, surrounded by single ticks). So...I would ask you to reconsider on this. And for each of the other edge cases, either choose to address them or not (we don't have to address all of them).

b-kelly:
@yellis I tend to prioritize user experience over all else. In this case the problem wasn't with the user's markup, but the auto-fencing feature itself. Since that feature was causing more issues than it solved (and was something I added because it already existed for rich-text and was simple to port), I have opted to remove it entirely.

I'd love to add it back in the future, but we'll need a much better solution than implementing a partial markdown parser to detect where it is and isn't valid to paste.

PR #237, opened 2023-01-30
fix(code-block): keep code-block and inline-code menu buttons side by side

giamir:
This is an attempt to solve the usability issue identified in this meta post: https://meta.stackoverflow.com/questions/422686/code-blocks-surrounded-by-triple-backtick-then-single-backtick

**Describe your changes**
- Replace old `Codeblock` (filled) icon with the new `CodeblockAlt` (outlined) icon
- Create a separate area in the editor menu for code formatting (see mockup)
- Add additional help text in the Code block tooltip (see mockup)

Mockup:
Screenshot 2023-02-03 at 14 15 07

**PR Checklist**

### DIFF
--- a/src/shared/localization.ts
+++ b/src/shared/localization.ts
@@ -19,7 +19,7 @@ export const defaultStrings = {
         bold: shortcut("Bold"),
         code_block: {
             title: shortcut("Code block"),
-            helpText: "Use for code that is more than one line",
+            help_text: "Use for code that is more than one line",
         },
         emphasis: shortcut("Italic"),
         heading: {
--- a/src/shared/localization.ts
+++ b/src/shared/localization.ts
@@ -19,7 +19,7 @@ export const defaultStrings = {
         bold: shortcut("Bold"),
         code_block: {
             title: shortcut("Code block"),
-            help_text: "Use for code that is more than one line",
+            help_text: "Multiline block of code with syntax highlighting",
         },
         emphasis: shortcut("Italic"),
         heading: {
--- a/src/shared/localization.ts
+++ b/src/shared/localization.ts
@@ -19,7 +19,7 @@ export const defaultStrings = {
         bold: shortcut("Bold"),
         code_block: {
             title: shortcut("Code block"),
-            help_text: "Multiline block of code with syntax highlighting",
+            description: "Multiline block of code with syntax highlighting",
         },
         emphasis: shortcut("Italic"),
         heading: {
--- a/src/shared/localization.ts
+++ b/src/shared/localization.ts
@@ -17,7 +17,10 @@ export const defaultStrings = {
     commands: {
         blockquote: shortcut("Blockquote"),
         bold: shortcut("Bold"),
-        code_block: shortcut("Code block"),
+        code_block: {
+            title: shortcut("Code block"),
+            helpText: "Use for code that is more than one line",
+        },
         emphasis: shortcut("Italic"),
         heading: {
             dropdown: shortcut("Heading"),
--- a/src/shared/localization.ts
+++ b/src/shared/localization.ts
@@ -29,7 +29,10 @@ export const defaultStrings = {
         help: "Help",
         horizontal_rule: shortcut("Horizontal rule"),
         image: shortcut("Image"),
-        inline_code: shortcut("Inline code"),
+        inline_code: {
+            title: shortcut("Inline code"),
+            description: "Single line code span for use within a block of text",
+        },
         kbd: shortcut("Keyboard"),
         link: shortcut("Link"),
         metaTagLink: shortcut("Meta tag"),
--- a/src/shared/menu/entries.ts
+++ b/src/shared/menu/entries.ts
@@ -304,21 +304,6 @@ export const createMenuEntries = (
                     "italic-btn"
                 ),
             },
-            {
-                key: "toggleCode",
-                richText: {
-                    command: toggleMark(schema.marks.code),
-                    active: markActive(schema.marks.code),
-                },
-                commonmark: inlineCodeCommand,
-                display: makeMenuButton(
-                    "Code",
-                    _t("commands.inline_code", {
-                        shortcut: getShortcut("Mod-K"),
-                    }),
-                    "code-btn"
-                ),
-            },
             addIf(
                 {
                     key: "toggleStrike",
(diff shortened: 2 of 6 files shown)


src/shared/menu/helpers.ts
```diff
@@ -225,14 +225,15 @@ export function addIf(item: MenuItem, flag: boolean): MenuItem {
...
+ * @param content Either a string containing the text to place in the button's title attribute
+ * or an object containing the title and helpText to be used in the hover tooltip
  * @param key A unique identifier used for identifying the command to be executed on click
  * @param cssClasses extra CSS classes to be applied to this menu icon (optional)
  * @internal
  */
 export function makeMenuButton(
     iconName: string,
-    title: string,
+    content: string | { title: string; helpText: string },
```
giamir, line 236: @b-kelly  I am not super happy with this change. I did not want to add a 5th argument to this function since the parameter list is already long enough. At the same time I did not want to change the signature for all function callers with an object (see the many callers in entries.ts file) so I settled with this union type.
Any suggestion?

src/shared/menu/helpers.ts
```diff
@@ -243,6 +244,13 @@ export function makeMenuButton(
         button.classList.add(...cssClasses);
     }
 
+    const title = typeof content === "string" ? content : content.title;
+    const helpText = typeof content === "string" ? null : content.helpText;
+
+    if (helpText) {
+        button.dataset.sTooltipHtmlTitle = `<p class="m0">${title}</p><p class="fs-caption fc-black-600 m0">${helpText}</p>`;
```
dancormier, line 251: Nit: spacing as-is looks a little tight in the context of the padded popover. I suggest adding a little space between lines.
```suggestion
        button.dataset.sTooltipHtmlTitle = `<p class="mb4">${title}</p><p class="fs-caption fc-black-600 m0">${helpText}</p>`;
```

src/shared/localization.ts
```diff
@@ -17,7 +17,10 @@ export const defaultStrings = {
     commands: {
         blockquote: shortcut("Blockquote"),
         bold: shortcut("Bold"),
-        code_block: shortcut("Code block"),
+        code_block: {
```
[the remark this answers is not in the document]
dancormier: I agree in general, but suggest `snake_case` in the instance of commands since that matches the [basic ProseMirror schema](https://prosemirror.net/docs/ref/#schema-basic).

dancormier (commented):
This PR does what it describes, but I have a nit and there's a few questions to be answered before a ✅ makes sense here.

---

### Potentially out of scope

I noticed some odd behavior when when clicking applying "toggleCode" on a selected code block:



We could consider this out of scope for the intent of this PR, but I figured I'd call it out.

src/shared/localization.ts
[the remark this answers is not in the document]
giamir: I might actually opening Pandora's box here:
- It looks like we are not consistent in the naming of schema nodes (at least [there is this exception](https://github.com/StackExchange/Stacks-Editor/blob/eb7e1c50204b3884ccefc1390942944845d0b126/src/rich-text/schema.ts#L429) that propagates in the localization file)
- In the [defaultStrings object](https://github.com/StackExchange/Stacks-Editor/blob/eb7e1c50204b3884ccefc1390942944845d0b126/src/shared/localization.ts#L16) we group things around the `commands` keyword although not everything is a command really. All are `menu_entries` though.

I think I will make a follow up PR with a refactor, trying to normalize a bit the conventions.
For this one I am not going to take any action apart from renaming `helpText` to `help_text` to follow the convention used by the majority of the items in the `defaultStrings` object.

giamir:
Thanks @dancormier for the review! 🙏

> I noticed some odd behavior when when clicking applying "toggleCode" on a selected code block

[That is behavior already existing today in the editor](https://editor.stackoverflow.design/) so as you suggested I will consider it out of scope for this PR.

If we believe that's a bug we should probably document it and also describe the expected behavior. 
@shanefoster @b-kelly @yellis Do you believe that is something that contributes to [this meta issue](https://meta.stackoverflow.com/questions/422686/code-blocks-surrounded-by-triple-backtick-then-single-backtick) as well? I would suggest to tackle the edge case Dan described only if the meta issue persists after this PR is released.

yellis:
> Do you believe that is something that contributes to [this meta issue](https://meta.stackoverflow.com/questions/422686/code-blocks-surrounded-by-triple-backtick-then-single-backtick) as well

I think that it should be treated as a separate case. It is something I have seen, and maybe folks who just dont know what they are doing are running into both of them, but doesn't really overlap with the core of what this is trying to accomplish.

src/shared/localization.ts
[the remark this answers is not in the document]
b-kelly: Ah yeah - if there are any items in there that are `camelCased`, then that's an oversight and should be remediated (follow-up PR). I don't recall the exact reason I made these `snake_case` to begin with, but I feel like it was probably a good reason. I'll think on it a bit and see if I can't remember.

src/shared/menu/helpers.ts
```diff
@@ -243,6 +244,13 @@ export function makeMenuButton(
         button.classList.add(...cssClasses);
     }
 
+    const title = typeof content === "string" ? content : content.title;
+    const helpText = typeof content === "string" ? null : content.helpText;
+
+    if (helpText) {
+        button.dataset.sTooltipHtmlTitle = `<p class="mb4">${title}</p><p class="fs-caption fc-black-600 m0">${helpText}</p>`;
```
b-kelly, line 251: Could be a potential XSS here. Should use the html formatter helper instead.

src/shared/menu/helpers.ts
b-kelly: I think this is an ok change overall. I'll look at it in code and see if I can't think of a better way to express that without having to update every caller

src/shared/localization.ts
```diff
@@ -17,7 +17,10 @@ export const defaultStrings = {
...
+            title: shortcut("Code block"),
+            help_text: "Use for code that is more than one line",
```
b-kelly, line 22: This reads a little weird to me. I'll update the text a bit

b-kelly (commented):
Looking good! I'm going to make a few small tweaks since you're out for the day and give the PR a merge

b-kelly (approved):
Made a few tiny tweaks (XSS, wording, added description for inline code as well).

LGTM, thanks for tackling this!

src/shared/menu/helpers.ts
giamir: I thought that translations won't be considered user generated content. I guess better be extra safe nevertheless. 🙂

src/shared/localization.ts
```diff
@@ -19,7 +19,7 @@ export const defaultStrings = {
         bold: shortcut("Bold"),
         code_block: {
             title: shortcut("Code block"),
-            help_text: "Use for code that is more than one line",
+            help_text: "Multiline block of code with syntax highlighting",
```
giamir: @b-kelly I am not against changing the text per se but in general I would recommend to double check with the author before doing so. Copy is coming from Design.
@shanefoster can you give us a thumbs up about this text change?

src/shared/menu/entries.ts
```diff
@@ -335,9 +335,12 @@ export const createMenuEntries = (
                 commonmark: inlineCodeCommand,
                 display: makeMenuButton(
                     "Code",
-                    _t("commands.inline_code", {
-                        shortcut: getShortcut("Mod-K"),
-                    }),
+                    {
```
giamir, line 338: @b-kelly I understand your good intentions here but we cannot take this kind of executive decisions after a group of designers have put an effort to analyse the story. I might even agree with you here but we need to make these suggestions transparent.

PR #241, opened 2023-03-10
fix(commonmark): allow shift-enter command

### LINKED ISSUE #240: Markdown Mode: Support using Shift + Enter to add newline
**Describe the bug**
Using <kbd>Shift</kbd> + <kbd>Enter</kbd> does not add a new line in Markdown mode (but does in Rich Text Mode)

**To Reproduce**

1. Go to editor in Markdown Mode
2. Hit <kbd>Enter</kbd> and correctly see new line
3. Hit <kbd>Shift</kbd> + <kbd>Enter</kbd> and you no longer get a new line

**Expected behavior**
This is how most editors in my experience behave, including the previous editor. In _some_ forms, some of the time, hitting <kbd>Enter</kbd> will submit the entire form, and so using Shift + Enter is a common keyboard usage that helps prevent unintended form submissions.

**Desktop**
 - OS: Windows
 - Browser Chrome
 - Version 110

KyleMit:
Closes #240

## **Describe your changes**

Adds a keybinding for `Shift-Enter` and reuses the existing command bound to `Enter`

## **PR Checklist**

^ Could use some help with the above.  Not sure we need to super robustly test since we're just exporting an existing command as a new key binding

## **Environment(s) tested**

- Device: Desktop
- OS: Windows
- Browser: Chrome
- Version: 110

## **Additional context**

The default functionality comes from [`prosemirror-commands/src/commands.ts`](https://github.com/ProseMirror/prosemirror-commands/blob/7d895cae5224ad6d5bd6e705878811673b0314f7/src/commands.ts#L691-L692)

```ts
export const pcBaseKeymap: {[key: string]: Command} = {
  "Enter": chainCommands(newlineInCode, createParagraphNear, liftEmptyBlock, splitBlock),
```

Which we import directly into our common mark key-bindings, along with any customizations we might like.  We could probably import the commands `chainCommands`, `newlineInCode`, `createParagraphNear`, `liftEmptyBlock`, `splitBlock` that all get run when the <kbd>Enter</kbd> is pressed, but I think it's easier to just take the built command that's part of the imported `baseKeymap`

## Old Behavior

[![shift-enter-old][1]][1]

## New Behavior

[![shift-enter-new][2]][2]

[1]: https://user-images.githubusercontent.com/4307307/224440792-863f5a47-301e-4fa3-b9b8-b1813cc5b557.gif
[2]: https://user-images.githubusercontent.com/4307307/224440877-d1722a2a-eb4f-4a8a-8402-e01408698d06.gif

### DIFF
--- a/src/commonmark/key-bindings.ts
+++ b/src/commonmark/key-bindings.ts
@@ -50,6 +50,7 @@ export function allKeymaps(parserFeatures: CommonmarkParserFeatures): Plugin[] {
         "Mod-,": subCommand,
         "Mod-.": supCommand,
         "Mod-'": kbdCommand,
+        "Shift-Enter": baseKeymap["Enter"],
         // selectAll selects the outermost node and messes up our other commands
         "Mod-a": selectAllTextCommand,
     });


giamir:
Thanks @KyleMit for the PR. ❤️ 
I will try to set some time aside to look into this tomorrow.
How urgent is it? Do you want us to cut a release straight after merge ideally?

KyleMit:
@giamir I wouldn't say urgent.  If we're planning on cutting a release in the next several weeks, then we can pull in then - if not, would be nice to have when we merge.  But this has existed on Prod since the inception of the editor and more a quality of life fix than an outright "bug"

giamir (approved):
LGTM. We don't explicitly add tests for keybindings as of today (we treat them a bit as configuration files).
It's interesting to notice that the only reason `Shift-Enter` is adding a new paragraph in rich text mode is because of a "side effect" of the `exitBlockCommand` ([code](https://github.com/StackExchange/Stacks-Editor/blob/main/src/rich-text/key-bindings.ts#L69)). 
I will merge this change now and cut a release soon - I will leave it to you to update Core then.
Thanks again for your contribution.

PR #244, opened 2023-04-10
fix(a11y): accessible mode switcher

dancormier:
Closes [Stacks-286](https://stackoverflow.atlassian.net/browse/STACKS-286)

**Describe your changes**

This PR updates the mode switcher to use the [Stacks `.s-btn-group--radio` variant syntax](https://stackoverflow.design/product/components/button-groups/#radio), which is keyboard navigable.

**PR Checklist**

**Environment(s) tested**

- Device: MacBook Pro
- OS: Mac OS 13.0.1
- Browser Firefox, Chrome

**Additional context**

This PR also remove auto-focusing the editor frame on mode switching for ergonomics/accessibility.

### DIFF
--- a/src/stacks-editor/editor.ts
+++ b/src/stacks-editor/editor.ts
@@ -345,13 +345,13 @@ export class StacksEditor implements View {
         const container = document.createElement("div");
         container.className = "flex--item d-flex ai-center ml24 fc-medium";
 
-        container.innerHTML = escapeHTML`<div class="s-btn-group__radio">
+        container.innerHTML = escapeHTML`<div class="s-btn-group s-btn-group--radio fw-nowrap">
     <input type="radio" name="mode-toggle-${this.internalId}"
         id="mode-toggle-rich-${this.internalId}"
-        class="js-editor-toggle-btn"
+        class="s-btn--radio js-editor-toggle-btn"
         data-mode="${EditorType.RichText}"
         ${richCheckedProp} />
-    <label class="s-btn s-editor-btn px6"
+    <label class="s-btn s-btn__muted s-btn__outlined s-editor-btn px6"
         for="mode-toggle-rich-${this.internalId}"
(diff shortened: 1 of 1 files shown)


giamir (approved):
LGTM Dan. Works great! 

I would like for us to add a small test to catch potential regressions in the future.
Disclaimer: I don't fully understand how the tests have been structured honestly in this repo.

We could extend or add a test in this e2e suite simulating the user clicking the tab and arrow functions to switch modes.
https://github.com/StackExchange/Stacks-Editor/blob/824437823fbf89e59b83367214f93d9660984653/test/rich-text/editor.e2e.test.ts#L37
The issue is that this behavior is also expected on the commonmark mode and in theory a test should be added here too:
https://github.com/StackExchange/Stacks-Editor/blob/main/test/commonmark/editor.e2e.test.ts

**Perhaps the best solution would be to refactor the `switchMode` helper function to use real user interactions (tab / arrow keys) instead of using JS (querySelector):
https://github.com/StackExchange/Stacks-Editor/blob/main/test/e2e-helpers.ts#L17-L27**

Let me know if you need my support for making those changes.
Thanks 🙂

dancormier:
> Perhaps the best solution would be to refactor the switchMode helper function to use real user interactions (tab / arrow keys) instead of using JS (querySelector): https://github.com/StackExchange/Stacks-Editor/blob/main/test/e2e-helpers.ts#L17-L27

I attempted to in https://github.com/StackExchange/Stacks-Editor/pull/244/commits/55639d3568d092b9d9a9de5ea89dc933f2726b7f but several tests now fail. It was stumping me since it seemed that the editor was in the appropriate mode for each test. @giamir mind lending your eyes to this or happy to sync on it to consider the most sensible way to test this.

giamir:
@dancormier I refactored a bit the test utils. Have a look and feel free to merge.
We should also cut a release for this issue and open a PR in core. 🙂
Thank you

PR #246, opened 2023-05-09
fix(a11y): remove illegal (and superfluous) aria role from the image uploader browse label

balpha:
TEAMS-9651, items 2005562 and 2005563

This was called out in an accessibility audit. A label can't have a button role, and it's not necessary anyway -- since it labels the "choose file" button, it's announced as a button even without this role. I confirmed this with NVDA/Chrome.


giamir (approved):
Thanks @balpha. LGTM.

PR #247, opened 2023-05-15
PoC for intercepting image uploads

giamir:
This PoC explore how we could leverage the existing `ImageUploadOptions["handler"]` to intercept image uploads and show a prompt to the user to swap images with detected text. 

## How to test:
- Start the webpack dev server `npm start`
- Visit `http://localhost:8080/intercept-image.html`
- Make sure you are on rich text mode
- Click on the image menu item
- Add an image
- See modal popping up

## Known issue
It is not possible to inject data to the editor directly from the handler
See issue in code: https://github.com/StackExchange/Stacks-Editor/compare/STACKS-310/intercept-image-flow-poc?expand=1#diff-be39fdbd3dacf49598cf929839b9ab9e1629658c2195c2d8b04826da45ca678cR176

## Possible solutions (need discovery)
- Hide the image menu item and build a brand new editor plugin to achieve exactly what needed
- Expand the `ImageUploadOptions` interface to support injecting data into the editor

PR #248, opened 2023-05-15
Image upload intercept plugin

b-kelly:
Very basic sample plugin that shows how one might listen to, intercept and handle editor events. PR built in response to suggestion in #247.


giamir:
Closing this PoC PR to avoid causing confusion.
Please consider https://github.com/StackExchange/Stacks-Editor/pull/249 as the single source of truth.

PR #249, opened 2023-05-16
feat(image-upload): dispatch event for intercepting image upload flow

giamir:
## Summary
Dispatch a custom event on image upload to allow plugin authors to pause the upload flow and decide to either abort or resume it based on async operations. 

## How to test
- Visit `http://localhost:8080/plugins.html` or the [PR deploy preview](https://deploy-preview-249--stacks-editor.netlify.app/plugins.html)
- Upload an image in rich-text or commonmark mode
- A confirmation dialog should appear saying that code was detected in the image
- Click ok and observe a codeblock being inserted in the editor as opposed to the image being uploaded

## Open Todos
- [x] review solution design 
- [x] write unit tests
- [x] cleanup demo plugin showcasing the feature

### DIFF
--- a/site/sample-plugins/code-detection.ts
+++ b/site/sample-plugins/code-detection.ts
@@ -1,0 +1,84 @@
+import { Plugin, PluginView } from "prosemirror-state";
+import { EditorView } from "prosemirror-view";
+import type { EditorPlugin } from "../../src";
+
+type InterceptImageUploadEvent = CustomEvent<{
+    file: File;
+    resume: (resume: boolean) => void;
+}>;
+
+class InterceptImageUploadView implements PluginView {
+    private listener: (e: InterceptImageUploadEvent) => void;
+
+    constructor(private view: EditorView) {
+        this.listener = (e: InterceptImageUploadEvent) =>
+            void (async () => {
+                // pause the regular flow so that the image is not uploaded
+                // we can resume it by calling e.detail.resume(true) later on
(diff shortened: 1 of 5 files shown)


yellis:
This looks great. Only thing that missing that we would probably need is the ability to inject a modal workflow into the place where you have the confirmation js. Basically, we will want to intercept the image, upload it to extract the code, and then show the code (with syntax highlighting) to the user before adding it (which can be done in a modal). If we can have some sort of a callback that we could hook into where we would put in a function that would open the modal, and then be able to either exit the process or give code to inject, would be ideal.

giamir:
@yellis Our recommendation would be for your team to own and author the `code-detection` plugin. The `code-detection.ts` file you see in this PR is just an example of how you could author the plugin and leverage the new event API we created in the `image-upload` core plugin.
Our team will be happy to support you to create the plugin in the core codebase. In this playground we wanted to keep things rather simple but swapping the confirmation dialog over a modal would be a pretty straightforward operation (it would just make the example a bit more verbose).
Essentially the only source code changes we plan to make on the editor are the one in the [`image-upload.ts`](https://github.com/StackExchange/Stacks-Editor/pull/249/files#diff-12ad4d07af5406ec4ec5e890af42301d3945df8ab50de929a9e3b8fce6d6040a) file.
Happy to discuss this further if necessary. 🙂

dancormier (approved):

PR #254, opened 2023-09-12
fix(a11y): prevent refocus of textarea on keyboard navigation

dancormier:
Closes [STACKS-383](https://stackoverflow.atlassian.net/browse/STACKS-383)

This PR ensures that the focused editor menu element retains focus when navigating via keyboard.

Previously, tabbing to a menu element then invoking it with <kbd>enter</kbd> would send focus to the "view" (aka the textarea). This would require those navigating with the keyboard to have to tab back to where they were to continue where they left off. This PR addresses this issue with the following changes:

1. Focus is no longer moved to the view unless editor menu item was clicked
2. The popover is only hidden if the popover menu item was clicked

### TODO
- [x] Add unit test
- [x] Add ticket(s) for fully accessible menu (see https://github.com/StackExchange/Stacks-Editor/pull/254#pullrequestreview-1624199129) https://stackoverflow.atlassian.net/browse/STACKS-402

### DIFF
--- a/src/shared/menu/plugin.ts
+++ b/src/shared/menu/plugin.ts
@@ -112,7 +112,22 @@ export class MenuView implements PluginView {
             const key = (target as HTMLElement).dataset.key;
 
             e.preventDefault();
+            const isMouseEvent = e.detail > 0; // See https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/detail
 
+            // Move focus to the editor exclusively for mouse clicks
+            // For keyboard events, we keep the focus on the menubar
+            if (isMouseEvent) {
+                // When the click is from a menuitem, hide the popover
+                if (target.getAttribute("role") === "menuitem") {
+                    const menuButton = (<HTMLElement>e.target).closest(
+                        '[data-controller="s-popover"]'
+                    );
+                    hidePopover(menuButton);
+                }
+
+                // leave the menubar and focus on the editor
+                view.focus();
+            }
             // Conditional added to only focus view on mouse events, so keyboard navigation remains intact
             // See https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/detail
             if (e.detail > 0) {
--- a/src/shared/menu/plugin.ts
+++ b/src/shared/menu/plugin.ts
@@ -112,7 +112,18 @@ export class MenuView implements PluginView {
             const key = (target as HTMLElement).dataset.key;
 
             e.preventDefault();
-            view.focus();
+
+            // Conditional added to only focus view on mouse events, so keyboard navigation remains intact
+            // See https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/detail
+            if (e.detail > 0) {
+                // Hide the menu popover if it's visible
+                const menu = (<HTMLElement>e.target).closest(".s-popover");
+                if (target.getAttribute("role") === "menuitem" && menu) {
+                    menu?.classList.remove("is-visible");
+                    target.setAttribute("aria-expanded", "false");
+                }
+                view.focus();
+            }
 
             const found = menuCommands.find((c) => c.key === key);
             const foundCommand = this.command(found);
@@ -319,7 +330,7 @@ export class MenuView implements PluginView {
         button.setAttribute("aria-controls", popoverId);
         button.setAttribute("data-action", "s-popover#toggle");
         button.setAttribute("data-controller", "s-tooltip");
-        button.setAttribute("role", "menuitem");
+        button.setAttribute("role", "menu");
         button.id = buttonId;
         button.dataset.key = entry.key;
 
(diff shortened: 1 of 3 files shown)


src/shared/menu/plugin.ts
```diff
@@ -112,7 +112,18 @@ export class MenuView implements PluginView {
             const key = (target as HTMLElement).dataset.key;
 
             e.preventDefault();
-            view.focus();
+
```
giamir, line 115:
```suggestion
            const isMouseEvent = e.detail > 0; // See https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/detail

            // Move focus to the editor exclusively for mouse clicks
            // For keyboard events, we keep the focus on the menubar
            if (isMouseEvent) {
                // When the click is from a menuitem, hide the popover
                if (target.getAttribute("role") === "menuitem") {
                    const menuButton = (<HTMLElement>e.target).closest(
                        '[data-controller="s-popover"]'
                    );
                    hidePopover(menuButton);
                }

                // leave the menubar and focus on the editor
                view.focus();
            }
```

The `aria-expanded` attribute is not in the `target` element (which is a menuitem in this case) . I think it's a better idea to handle popover management directly with Stacks APIs. `import { hidePopover } from "@stackoverflow/stacks";`

The suggested snippet worked as expected in my local environment.

giamir (commented):
Thanks @dancormier for this small fix (see comment in the code - there is a small bug). 
We can add a unit test and we should be good to move on with this story.

## Sidenote (More a11y issues)
While reviewing this fix I spent a good amount of time reading [WAI authoring practices for menu and menubar](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/). I think that what we have now is still not fully accessible. 
Here are the issues I understood still affect our implementation:
- When a menu opens, keyboard focus should be placed on the first item (at the moment the focus stays on the menu button)
- `Tab` and `Shift + Tab` should NOT move focus among `menuitem`s instead they should move to the next/previous focusable item which is not a `menuitem` and close the menu if it was open.
- `Arrow Up` and `Arrow Down` should be used instead to loop through the `menuitem`s (focus should be trapped)
- When `Enter` (and optionally `Space`) is pressed and user is focused on a `menuitem` the item should activate and the menu should close (at the moment the menu remains open)

I think we could consider these a11y issues outside the scope of this fix (since they were not in the report). That said I would not be surprised if we will get a follow up report with these issues listed. The minimum we should do is to create a ticket for our backlog and in my opinion we should try to work on that ticket before the enterprise release too to be on the safe side.

What are your thoughts @dancormier?

I am also tagging @AvogadroSG1 for visibility/3rd opinion.

dancormier:
> Thanks @dancormier for this small fix (see comment in the code - there is a small bug). We can add a unit test and we should be good to move on with this story.

I'll add a unit test and merge it in. It's not perfect, but it accomplishes the goal at hand and is an improvement over what we had.

> ## Sidenote (More a11y issues)
> 
> While reviewing this fix I spent a good amount of time reading [WAI authoring practices for menu and menubar](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/). I think that what we have now is still not fully accessible. Here are the issues I understood still affect our implementation:
> 
>     * When a menu opens, keyboard focus should be placed on the first item (at the moment the focus stays on the menu button)
> 
>     * `Tab` and `Shift + Tab` should NOT move focus among `menuitem`s instead they should move to the next/previous focusable item which is not a `menuitem` and close the menu if it was open.
> 
>     * `Arrow Up` and `Arrow Down` should be used instead to loop through the `menuitem`s (focus should be trapped)
> 
>     * When `Enter` (and optionally `Space`) is pressed and user is focused on a `menuitem` the item should activate and the menu should close (at the moment the menu remains open)

Thank you for this research. I kinda took the ticket/report at face value and "fixed" the issue, but the approach you found in your research makes total sense. When I was working on this, I even had the thought about arrow navigation making sense (I thought it would be nice if it kinda worked like navigating a radio group), but just left it as-is.

Out of curiosity (and proximity), I tested out the `⋮` button on the top-right corner of these PR comments to see how they work. GitHub is not without its accessibility issues but I figured they'd meet MSFT's standards considering they're GitHub's parent company. What they do is _close_ to what WCAG describes, with the exceptions that:

- Invoking the menu _does not_ focus the first item
- <kbd>Tab</kbd> navigates the `menuitem`s in addition to the arrows



For the record, I'm not suggesting we follow GitHub's lead here. I figured I'd mention it as a curiosity here.

> I think we could consider these a11y issues outside the scope of this fix (since they were not in the report). That said I would not be surprised if we will get a follow up report with these issues listed. The minimum we should do is to create a ticket for our backlog and in my opinion we should try to work on that ticket before the enterprise release too to be on the safe side.

I agree. I think we can consider this as accomplishing the goal and resolving the _stated issue_. There's room for improvement, but this is moving in the right direction. Related, I've yet to see a concrete standard we're being held to by the third-party requesting the fix. The best I've heard is "we need to be able to show we're improving accessibility" which I think this accomplishes. We can prioritize other improvements that get us closer to something like WCAG AA adherence outside of this PR. For now, I'll write up some tickets.

Thanks @giamir for the review and all the information!

PR #260, opened 2023-10-30
fix(a11y): use disabled attribute instead of .is-disabled

dancormier:
**Describe your changes**

This change came from some issues I noticed while reviewing https://github.com/StackExchange/Stacks-Editor/pull/259. I took this opportunity to replace uses of `.is-disabled` with the `disabled` attribute which is ensures that disabled editor buttons act in accordance with our expectation of disabled UI. It has the added benefit of allowing us to some custom CSS, since Stacks buttons apply the expected styling on disabled buttons by targeting the `disabled` attribute.

**PR Checklist**

@giamir I've opened this PR because I think it is fine to be reviewed, but no rush on this. It's just some work I performed in the process of reviewing https://github.com/StackExchange/Stacks-Editor/pull/259 and I felt was worth capturing.

### DIFF
--- a/src/shared/menu/plugin.ts
+++ b/src/shared/menu/plugin.ts
@@ -46,7 +46,6 @@ export class MenuView implements PluginView {
     private readonly: boolean;
     private editorType: EditorType;
 
-    static disabledClass = "is-disabled";
     static activeClass = "is-selected";
     static invisibleClass = "d-none";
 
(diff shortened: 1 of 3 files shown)


giamir (approved):
Nice work! ❤️

Somehow I was expecting this to be a bigger refactor. I think this is small enough to get merged and be released alongside the HC fix we have done in the other PR.

Thanks @dancormier

PR #262, opened 2024-01-25
fix(autoformatting): prevent autoformatting when the first underscore is preceded by a non-whitespace character

giamir:
[STACKS-520](https://stackoverflow.atlassian.net/browse/STACKS-520)

This PR is fixing an issue where Underscore Emphasis Autoformatting was triggered incorrectly. 
See the how to test section for more context.

The solution is leveraging a negative lookahead operator to make sure the first underscore is NOT preceded by a non-whitespace character.

# How to test
- Open the editor instance in rich text mode
- Type manually (not copy pasting) `text_text_` (or `text__text__`)
- Observe auto formatting NOT kicking in and rewriting the above to text<em>text</em> (or text<strong>text</strong>)

### DIFF
--- a/src/rich-text/inputrules.ts
+++ b/src/rich-text/inputrules.ts
@@ -17,7 +17,7 @@ export const emphasisRegexAsterisk = /\*([^*\s]([^*])*[^*\s]|[^*\s])\*$/;
 // matches: __some text__, but not __ text __
 const strongRegexUnderscores = /__(\S(?:|.*?\S))__$/;
 // matches: _some text_, but not __text_ or _ text _
-const emphasisRegexUnderscore = /_([^_\s]([^_])*[^_\s]|[^_\s])_$/;
+const emphasisRegexUnderscore = /(?<!\S)_([^_\s]([^_])*[^_\s]|[^_\s])_$/;
 // matches: [ *any* thing ]( any thing )
 const linkRegex = /\[(.+)\]\((.+)\)$/;
 
--- a/src/rich-text/inputrules.ts
+++ b/src/rich-text/inputrules.ts
@@ -15,7 +15,7 @@ const strongRegexAsterisks = /\*\*(\S(?:|.*?\S))\*\*$/;
 // matches: *some text*, but not **text* or * text *
 export const emphasisRegexAsterisk = /\*([^*\s]([^*])*[^*\s]|[^*\s])\*$/;
 // matches: __some text__, but not __ text __
-const strongRegexUnderscores = /__(\S(?:|.*?\S))__$/;
+const strongRegexUnderscores = /(?<!\S)__(\S(?:|.*?\S))__$/;
 // matches: _some text_, but not __text_ or _ text _
 const emphasisRegexUnderscore = /(?<!\S)_([^_\s]([^_])*[^_\s]|[^_\s])_$/;
 // matches: [ *any* thing ]( any thing )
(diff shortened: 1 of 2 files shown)


src/rich-text/inputrules.ts
```diff
@@ -17,7 +17,7 @@ export const emphasisRegexAsterisk = /\*([^*\s]([^*])*[^*\s]|[^*\s])\*$/;
 // matches: __some text__, but not __ text __
 const strongRegexUnderscores = /__(\S(?:|.*?\S))__$/;
 // matches: _some text_, but not __text_ or _ text _
-const emphasisRegexUnderscore = /_([^_\s]([^_])*[^_\s]|[^_\s])_$/;
+const emphasisRegexUnderscore = /(?<!\S)_([^_\s]([^_])*[^_\s]|[^_\s])_$/;
```
hamvocke, line 20: would we also need to apply this to the double-underscore version above?

right now, the same bug would appear with something like `text__no__match` - does that matter?

giamir: We can do that. There might be cases where people name their DB tables with a double underscore. 😀
Honestly I was thinking that maybe we could apply this non-whitespace negative lookahead for all the rules but then I was like if somebody has a problem with how things work they will tell us. No need for premature optimization.

giamir: Addressed by https://github.com/StackExchange/Stacks-Editor/pull/262/commits/31d00a72f862432dabb59240976d3d27ff579f5a

dancormier (approved):

src/rich-text/inputrules.ts
```diff
@@ -15,7 +15,7 @@ const strongRegexAsterisks = /\*\*(\S(?:|.*?\S))\*\*$/;
 // matches: *some text*, but not **text* or * text *
 export const emphasisRegexAsterisk = /\*([^*\s]([^*])*[^*\s]|[^*\s])\*$/;
 // matches: __some text__, but not __ text __
-const strongRegexUnderscores = /__(\S(?:|.*?\S))__$/;
+const strongRegexUnderscores = /(?<!\S)__(\S(?:|.*?\S))__$/;
```
makyen, line 18: Just FYI: For performance reasons it's better to put the lookaround deeper into the regular expression, particularly not as the first thing in the regex.

Basically `/__(?<!\S__)(\S(?:|.*?\S))__$/` will be significantly more performant than `/(?<!\S)__(\S(?:|.*?\S))__$/`. This is because then matching the lookbehind is only evaluated once the `__` matches instead of on *every single character*.

The cheapest thing for a regular expression engine to do is just testing to see if the current character is a fixed character, such as is the current character being a "_". Putting the lookbehind first means that the regex engine does more than double the work on nearly every character (i.e., on every character that's preceded by a `\S` both the preceding character and the current character are tested instead of testing just the current character; and that doesn't even get into it being more expensive to enter and leave a lookbehind and more expensive to compare to a set, which requires comparing against multiple values, than to compare to the single character "_").

What the performance impact is of such a change will depend on the specific regular expression engine, the regular expression, and the text being tested. In general, I've seen it be as little as a 10% improvement to as much as close to 90%+ of the computation. In the real world, we can see 10s of seconds reduction in regex search times on metasmoke by using similar changes in the regex used for the search. 

Obviously, the same performance consideration applies to `emphasisRegexUnderscore` below.

I'm not saying "this needs to change". I'm just indicating something that should be generally considered when creating a regular expression.

giamir: Thanks for this input. Out of my personal curiosity I have applied the suggestion you made and compare execution timings against our unit test suite. While I believe your point is valid I could not see any relevant improvements in our context at least so we won't retroactively make changes here but it's something we will keep in mind going forward. 
Thanks again.

PR #267, opened 2024-01-31
fix(controls): improve focus style to meet WCAG 2.2 AA

dancormier:
Closes [STACKS-542](https://stackoverflow.atlassian.net/browse/STACKS-542)

This PR updates focus styles for buttons that are editor controls.

> **Note**
> This PR removes styling for `.s-btn-group__radio`. This element no longer exists (it's now `.s-btn-group--radio`).

> ~**Note**~
> ~This output of this PR may look incorrect until the @StackExchange/stacks dependency is updated beyond `2.0.9`.~

### DIFF
--- a/src/styles/custom-components.css
+++ b/src/styles/custom-components.css
@@ -6,53 +6,6 @@
     }
 }
 
-/* TODO UPSTREAM */
-.s-btn-group__radio {
-    display: flex;
-}
-
-/* TODO altered from upstream .s-btn-group>.s-btn */
-.s-btn-group__radio .s-btn:not(:last-of-type) {
-    margin-right: -1px;
-}
-
-.s-btn-group__radio .s-btn:not(:first-of-type):not(:last-of-type) {
-    border-radius: 0;
-}
-
-.s-btn-group__radio .s-btn:first-of-type:not(:only-of-type) {
-    border-bottom-right-radius: 0;
-    border-top-right-radius: 0;
-}
-
-.s-btn-group__radio .s-btn:last-of-type:not(:only-of-type) {
-    border-bottom-left-radius: 0;
-    border-top-left-radius: 0;
-}
-
-.s-btn-group__radio input[type="checkbox"],
-.s-btn-group__radio input[type="radio"] {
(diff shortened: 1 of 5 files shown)


giamir:
@dancormier What do you think of adding `@stackoverflow/stacks` in the peerDependencies field of the package.json of this library? 

```
"peerDependencies": {
    "@stackoverflow/stacks": "^2.1.0",
    "highlight.js": "^11.6.0"
}
```

[This reference](https://github.com/StackExchange/Stacks-Editor/blob/main/package.json#L89) will then be moved to the devDependencies object the same way we do for `highlight.js`.

I am suggesting this because the release containing the changes on this PR (probably `v0.10.0`) will rely on the styles of Stacks Classic v2.1.0 or above being rendered alongside the editor. Consumers failing to do so will potentially get a half broken editor. I also think that dependabot (for example in Core) will be smart enough and attempt to bump both libraries at once (worst case it will correctly fail CI when trying to run npm i updating exclusively the editor).

giamir:
> @dancormier What do you think of adding @stackoverflow/stacks in the peerDependencies field of the package.json of this library?

I took the liberty to cut a stacks classic release candidate and also add 2.1.0 as peer deps in this PR. 
I hope that's ok.

https://github.com/StackExchange/Stacks-Editor/pull/267/commits/4af3721add90712a4c09033c94238afe01c8c996

src/styles/custom-components.css
```diff
@@ -130,6 +96,11 @@
     right: var(--su2);
 }
 
+.s-btn-group--radio .s-btn.s-editor-btn {
```
giamir, line 99: I understand you want to change `.s-btn-group--radio` with `.s-btn-group__radio` as part of [this PR](https://github.com/StackExchange/Stacks/pull/1346). For now though this is correct.

giamir (approved):
Work as advertised. Great work @dancormier 

I am approving but let's remember we need to change stacks classic dep to 2.1.0 before merging and then cut the release.

CGuindon (approved):

PR #298, opened 2024-03-26
feat(list): enhance rich text list editing experience

giamir:
This PR enhance the editor rich text experience when editing lists. 
For more context about the issue see also [this document](https://docs.google.com/document/d/1rO6uRFIfELovhD5AAZRN-DuqV0JAUNhYhA7o1vjTLNc/edit?usp=sharing).

## How to test
- Go to the [empty editor page](https://deploy-preview-298--stacks-editor.netlify.app/empty) and switch to rich text mode

Part 1:
- Write at least 3 items in 3 different rows (e.g. `Item 1` then `Enter`, `Item 2` then `Enter` and so on)
- Select all the 3 items 
- Click on the bullet list menu button (or shortcut `Meta+u`)
- Observe the items becoming a bulleted list of 3
- Click on the bullet list menu button again  (or shortcut `Meta+u`)
- Observe the items being removed from the list
- Repeat the above steps with the ordered list button (shortcut `Meta+o`)

Part 2:
- Add in the editor a bullet list with 3 items
- Select the 2nd item
- Click on the bullet list menu button
- Observe the second item being removed from the list
- Click on the bullet list menu button again
- Observe the item being rejoined in a single list

All tests and linters should pass.

[STACKS-534](https://stackoverflow.atlassian.net/browse/STACKS-534)

### DIFF
--- a/config/jest-unit.config.js
+++ b/config/jest-unit.config.js
@@ -7,7 +7,10 @@ module.exports = {
     },
     rootDir: "../",
     testPathIgnorePatterns: ["/node_modules/", String.raw`\.e2e\.test`],
-    setupFilesAfterEnv: ["<rootDir>/test/matchers.ts"],
+    setupFilesAfterEnv: [
+        "<rootDir>/test/setup.ts",
+        "<rootDir>/test/matchers.ts",
+    ],
     transform: {
         "^.+\\.ts$": [
             "ts-jest",
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,6 +27,7 @@
                 "prosemirror-schema-list": "^1.3.0",
                 "prosemirror-state": "^1.4.3",
                 "prosemirror-transform": "^1.8.0",
+                "prosemirror-utils": "^1.2.1-0",
                 "prosemirror-view": "^1.33.1"
             },
             "devDependencies": {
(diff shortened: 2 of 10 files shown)


dancormier (approved):
This works as expected (with one _very_ minor quirk noted below) and the code seems reasonable. Thanks for taking care of this @giamir.

---

#### Noting one minor quirk

> [!NOTE]
> For the record, I don't think this behavior is significant or worth handling, so I don't suggest any changes to this PR.

I took the testing instructions in the PR description very literally and ran into an unexpected behavior. 

I added three items (and nothing else), I then selected them all with either `⌘ + a` or by clicking and dragging across all of them, and then entered either `⌘ + u` or clicked the bullet list button. I then entered either `⌘ + u` or clicked the bullet list button again and saw the bullet list get nested inside a new bullet list.



This only seems to happen when _all_ of the contents of the editor are selected. I imagine this circumstance is probably rare and is not an edge case worth handling, but felt like it was worth noting.

giamir:
> This only seems to happen when all of the contents of the editor are selected. I imagine this circumstance is probably rare and is not an edge case worth handling, but felt like it was worth noting.

I noticed this as well. This happen only if you select **all** the content. I agree with you we can consider this a rare edge case for now. Also I noticed that similar odd issues also happen with other commands when selecting **all** the content. Probably something that should be tackled in a separate PR anyway (not related to these changes).

Thanks for the review. 🎉

PR #304, opened 2024-04-12
feat: allow passthrough of accepted image file types to display in image upload UI

kristinalustig:
As discussed, this PR introduces a method by which users can pass through a list of strings representing different allowed media types. It then constructs an appropriate string listing out those media types and puts them into the image uploader dialog. Note that I am not sure what types of tests would be appropriate for this sort of change, if any, so if you think this change is large enough to require one, please let me know and I will update.

**PR Checklist**

**Additional context**

For reference, this pairs with SO codebase changes PR # 19347 - this one has to be in first so the build doesn't fail for SO.

### DIFF
--- a/src/shared/prosemirror-plugins/image-upload.ts
+++ b/src/shared/prosemirror-plugins/image-upload.ts
@@ -311,22 +311,13 @@ export class ImageUploader extends PluginInterfaceView<
     }
 
     getAcceptedFileTypesString(types: string[]): string {
-        if (types?.length === 0) {
+        if (types.length === 0) {
             return "";
         }
 
         let uploadCaptionString = types[0].replace(/image\//g, "");
         if (types.length > 1) {
-            uploadCaptionString = types
-                .map((type, i) => {
-                    return i === types.length - 1
-                        ? type
-                        : i === types.length - 2
-                          ? `${type}, or`
-                          : `${type},`;
-                })
-                .join(" ")
-                .replace(/image\//g, "");
+            uploadCaptionString = types.join(", ").replace(/image\//g, "");
         }
 
         return uploadCaptionString;
--- a/src/shared/prosemirror-plugins/image-upload.ts
+++ b/src/shared/prosemirror-plugins/image-upload.ts
@@ -57,6 +57,10 @@ export interface ImageUploadOptions {
      * If true, allow users to add images via an external url
      */
     allowExternalUrls?: boolean;
+    /**
+     * A string-based list that takes any number of file type names (e.g. bmp, gif, etc.) as input and uses them to populate help text in image uploader
+     */
+    acceptedFileTypes?: string[];
 }
 
 /**
@@ -140,6 +144,9 @@ export class ImageUploader extends PluginInterfaceView<
         this.uploadField.accept = "image/*";
         this.uploadField.multiple = false;
         this.uploadField.id = "fileUpload" + randomId;
+        const acceptedFileTypeString = this.createAcceptedFileTypeString(
+            uploadOptions.acceptedFileTypes
+        );
 
         // TODO i18n
         this.uploadContainer.innerHTML = escapeHTML`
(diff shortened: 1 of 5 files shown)


dancormier (changes requested):
Thanks for taking this on @kristinalustig!

We currently have a disconnect between the input and validation. Currently, we set the `accept` attribute on the input to `image/*` (so, **any** [image type](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types)). When the user tries to upload something else, we show a validation message  with `jpeg, png, gif` hardcoded:

image

In this PR, we don't address that but we are adding a field to pass file types. Right now, this is only used to generate caption text in the UI, but I think it can be used to populate the `accept` attribute on the input and ensure the validation message lines up with the file types we're accepting.

I've created https://github.com/StackExchange/Stacks-Editor/pull/305 targeting this PR as a proposal for a slightly different approach to what's in this PR. I _think_ it matches the mocks/requirements and just needs a test or two added to `image-upload.test.ts`. Check it out when you get a chance and we'll see if this is a worthwhile approach. I'm happy to chat or pair with you on Monday to go over any details if you'd like.

dancormier (commented):
I made some updates to how we handle the acceptedFileTypes option and how the they're displayed when rendered as text. I think this is in a good place to get merged.

Since I had a hand in much of these changes, I'd like to have @giamir give this a quick look just to make sure that I didn't do anything too funky and that the tests are sufficient.

src/shared/prosemirror-plugins/image-upload.ts
```diff
@@ -286,6 +310,41 @@ export class ImageUploader extends PluginInterfaceView<
...
+    getAcceptedFileTypesString(types: string[]): string {
+        if (types?.length === 0) {
+            return "";
+        }
+
+        let uploadCaptionString = types[0].replace(/image\//g, "");
+        if (types.length > 1) {
+            uploadCaptionString = types
+                .map((type, i) => {
+                    return i === types.length - 1
```
giamir, line 322: I personally find nested ternary operators far more difficult to read than nested ifs. I would probably just write `if`s in here even if it's a bit more verbose.

src/shared/prosemirror-plugins/image-upload.ts
```diff
@@ -286,6 +310,41 @@ export class ImageUploader extends PluginInterfaceView<
         event.stopPropagation();
     }
 
+    getAcceptedFileTypesString(types: string[]): string {
+        if (types?.length === 0) {
```
giamir, line 314: Can types ever be undefined? Doesn't look like it from the function signature.

src/shared/prosemirror-plugins/image-upload.ts
```diff
@@ -286,6 +310,41 @@ export class ImageUploader extends PluginInterfaceView<
...
+        }
+
+        let uploadCaptionString = types[0].replace(/image\//g, "");
+        if (types.length > 1) {
+            uploadCaptionString = types
+                .map((type, i) => {
+                    return i === types.length - 1
+                        ? type
+                        : i === types.length - 2
+                          ? `${type}, or`
```
giamir, line 325: I wonder if we need this extra complication here. I would personally challenge the requirement and ask design if we can drop the or here and just use commas as separators. On top of that `or` is not the same in every language and should not be hardcoded.

giamir: Maybe @kristinalustig can double check for us

giamir (changes requested):
Overall looks good. Great job. 👏 

I left a couple of small requests. One is challenging the product/design requirements to add an `or` (which is currently hardcoded). @kristinalustig you can explain to your designer, PM that the juice is likely not worth the squeeze for that `or`. Let us know if you need support in that conversation. 🙂

Tests are probably good enough I would say Dan. 👍

dancormier (approved):
@kristinalustig I removed the `getAcceptedFileTypesString` function since it was no only joining strings and removing `image/` in two spots. All tests pass and this looks good to me 🎉

giamir (approved):

giamir:
@kristinalustig we can merge this, do you want us to cut a new editor release with the changes straight away? @dancormier or I can do that instead of waiting till next week's "regular" release cycle. Let us know. 

Thank you for contributing to the editor codebase. 🎉

kristinalustig:
@giamir Ideally we can get it out sooner rather than later, but if that is a huge hassle it should be okay to wait until next week.

giamir:
> @giamir Ideally we can get it out sooner rather than later, but if that is a huge hassle it should be okay to wait until next week.

No, either @dancormier or myself will cut a release today or tomorrow. No reason to wait.

PR #305, opened 2024-04-12
Dcormier/dynamic accepted file types

dancormier:
description tbd

### DIFF
--- a/src/shared/prosemirror-plugins/image-upload.ts
+++ b/src/shared/prosemirror-plugins/image-upload.ts
@@ -131,6 +131,7 @@ export class ImageUploader extends PluginInterfaceView<
         super(INTERFACE_KEY);
 
         const randomId = generateRandomId();
+        const acceptedFileTypes = uploadOptions.acceptedFileTypes || [];
         this.isVisible = false;
         this.uploadOptions = uploadOptions;
         this.validateLink = validateLink;
@@ -143,7 +144,7 @@ export class ImageUploader extends PluginInterfaceView<
         this.uploadField = document.createElement("input");
         this.uploadField.type = "file";
         this.uploadField.className = "js-image-uploader-input v-visible-sr";
-        this.uploadField.accept = uploadOptions.acceptedFileTypes.join(", ");
+        this.uploadField.accept = acceptedFileTypes?.join(", ");
         this.uploadField.multiple = false;
         this.uploadField.id = "fileUpload" + randomId;
 
@@ -184,9 +185,8 @@ export class ImageUploader extends PluginInterfaceView<
         // add the caption element to the cta container
         const ctaContainer =
             this.uploadContainer.querySelector(".js-cta-container");
-        const acceptedFileTypesString = this.getAcceptedFileTypesString(
-            this.uploadOptions.acceptedFileTypes
-        );
+        const acceptedFileTypesString =
+            this.getAcceptedFileTypesString(acceptedFileTypes);
 
         if (acceptedFileTypesString) {
             const breakEl = document.createElement("br");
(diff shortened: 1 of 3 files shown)


src/shared/prosemirror-plugins/image-upload.ts
```diff
@@ -297,19 +307,36 @@ export class ImageUploader extends PluginInterfaceView<
...
+        if (JSON.stringify(acceptedTypes) !== JSON.stringify(defaultTypes)) {
+            if (acceptedTypes.length > 1) {
+                acceptedTypes[acceptedTypes.length - 1] =
+                    "or " + acceptedTypes[acceptedTypes.length - 1];
             }
-        });
-        return acceptedTypesString;
+            uploadCaptionString = acceptedTypes
+                .join(", ")
+                .replace(/image\//g, "");
```
kristinalustig, line 323: Why even have the `image` bit if we're just removing it any time we're displaying it? In this context it will always be an image.

src/shared/prosemirror-plugins/image-upload.ts
```diff
@@ -411,7 +438,12 @@ export class ImageUploader extends PluginInterfaceView<
                 return;
             case ValidationResult.InvalidFileType:
                 this.showValidationError(
-                    _t("image_upload.upload_error_unsupported_format")
+                    _t("image_upload.upload_error_unsupported_format", {
```
kristinalustig, line 441: I definitely missed this validation bit for the updated string, thank you

kristinalustig (commented):
I mostly just have a few questions for my own education/curiosity but I do think these updates meet our requirements. Once we've gone through those we can merge this into my original PR.

src/shared/prosemirror-plugins/image-upload.ts
dancormier: We're removing `image/` when we're _displaying_ it, but not always when we're using it. I figured it made the most sense to stick to the [MIME type](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types) to imply the connection to the file input's `accept` attribute. I'd be fine with inverting this if you think the API would be simpler to instead have the consumer provide a list of file extensions.

Let me know what you think and I'd be happy to make changes if you'd like.

src/shared/prosemirror-plugins/image-upload.ts
dancormier: No problem!

src/shared/prosemirror-plugins/image-upload.ts
giamir: > I figured it made the most sense to stick to the [MIME type](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types) to imply the connection to the file input's accept attribute.

FWIW sticking to the MIME type is a good call in my opinion. It will be easier for us to convey to the consumers how this option field should be populated - Furthermore people familiar with the HTML spec of `input type="file"` will have pretty much a zero learning curve. 🙂

src/shared/prosemirror-plugins/image-upload.ts
```diff
@@ -155,11 +155,7 @@ export class ImageUploader extends PluginInterfaceView<
             <div class="fs-body2 p12 pb0 js-cta-container">
                 <label for="${this.uploadField.id}" class="d-inline-flex f:outline-ring s-link js-browse-button" aria-controls="image-preview-${randomId}">
                     Browse
-                </label>, drag & drop<span class="js-external-url-trigger-container d-none">, <button type="button" class="s-btn s-btn__link js-external-url-trigger">enter a link</button></span>, or paste an image. 
-                <div class="fc-light fs-caption d-flex gsx gs4">
-                    <div class="flex--item">${acceptedFileTypeString}</div>
-                    <div class="flex--item">(Max size 2 MiB)</div>
-                </div>
+                </label>, drag & drop<span class="js-external-url-trigger-container d-none">, <button type="button" class="s-btn s-btn__link js-external-url-trigger">enter a link</button></span>, or paste an image.
```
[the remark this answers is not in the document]
kristinalustig: lol thanks github!

PR #310, opened 2024-04-24
fix(image-upload): increase text contrast for accepted files string to meet accessibility standards

kristinalustig:
**Describe your changes**

Updating the text contrast for the accepted file types string so that it meets accessibility standards for contrast.

**PR Checklist**

**Additional context**

Matches a change in SO editor.

### DIFF
--- a/src/shared/prosemirror-plugins/image-upload.ts
+++ b/src/shared/prosemirror-plugins/image-upload.ts
@@ -313,7 +313,7 @@ export class ImageUploader extends PluginInterfaceView<
 
     getCaptionElement(text: string): HTMLElement {
         const uploadCaptionEl = document.createElement("span");
-        uploadCaptionEl.className = "fc-light fs-caption";
+        uploadCaptionEl.className = "fc-black-500 fs-caption";
 
         let captionText = "(Max size 2 MiB)";
         if (text) {


giamir:
@kristinalustig the color contrast for the "caption" element is already passing our accessibility standard. This would be a cosmetic change rather than an accessibility fix. 

Before (`black-400` - APCA Lc76)

Screenshot 2024-04-29 at 11 18 23

After (`black-500` - APCA Lc94)
Screenshot 2024-04-29 at 11 17 37

Lc76 already passes our Lc60 target for text.

@kristinalustig Could you add a reference to a ticket? I believe a designer was consulted for making this change.
If not perhaps @CGuindon could give us the thumbs up on the change.

Thank you.

CGuindon:
Yup, `black-400` passes on our background and I would actually prefer to keep it at 400 instead of changing to 500. That way we only have 2 different shades of black in that small area instead of 3 (primary and secondary text). @kristinalustig

giamir:
@kristinalustig based on @CGuindon's last comment can we close this PR?

kristinalustig:
@giamir whoops, sorry for the delayed response, this PR got lost in the shuffle. Looks like we don't have to make this change, then. It's fine to close this PR. thanks @CGuindon for weighing in.

PR #352, opened 2024-09-20
feat: Invert button group icons for better contrast

abovedave:
Closes internal ticket \[STACKS-555\]. 

Currently the [button groups](https://stackoverflow.design/product/components/button-groups) in Stacks make the text bold on selected state. However the editor uses icons in place of text, which we obiously can't easily 'bolden'.

existing

This PR adds some override CSS to button groups for the editor to invert the icon against a darker background, making it much more accessible.

proposed

### DIFF
--- a/src/styles/custom-components.css
+++ b/src/styles/custom-components.css
@@ -59,12 +59,12 @@
     color: var(--black-600) !important;
 }
 
-.s-btn-group .s-btn--radio:checked + .s-btn {
+.s-btn-group .s-btn--radio:checked + .s-editor-btn {
     background: var(--black-500) !important;
     color: var(--white) !important;
 }
 
-.s-btn-group .s-btn--radio:focus-visible + .s-btn {
+.s-btn-group .s-btn--radio:focus-visible + .s-editor-btn {
     box-shadow:
         0 0 0 var(--su-static2) var(--focus-neutral),
         0 0 0 var(--su-static4) var(--focus-theme);
--- a/src/styles/custom-components.css
+++ b/src/styles/custom-components.css
@@ -65,7 +65,7 @@
 }
 
 .s-btn-group .s-btn--radio:focus-visible + .s-btn {
-    box-shadow: 
+    box-shadow:
         0 0 0 var(--su-static2) var(--focus-neutral),
         0 0 0 var(--su-static4) var(--focus-theme);
 }
--- a/src/styles/custom-components.css
+++ b/src/styles/custom-components.css
@@ -64,6 +64,12 @@
     color: var(--white) !important;
 }
 
+.s-btn-group .s-btn--radio:focus-visible + .s-btn {
+    box-shadow: 
+        0 0 0 var(--su-static2) var(--focus-neutral),
+        0 0 0 var(--su-static4) var(--focus-theme);
+}
+
 /*
     NOTE: These focus-visible styles exist to override the Stacks styles for editor buttons.
     Since the editor buttons require the focus rings to be applied outside of the button,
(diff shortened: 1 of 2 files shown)


CGuindon:
@abovedave How to I test stuff in the editor repo? I noticed I can't add the "PR environment" label?

abovedave:
@CGuindon added the deploy preview, I think it didn't do automatically as the PR is from my personal account.

abovedave:
@CGuindon good catch. We could remove the inset on the focus?

Screenshot 2024-09-24 at 14 33 20

CGuindon:
@abovedave I'd be ok with that for this type of icon only button group edge case. Would still love a gut check from @dancormier though.

CGuindon:
@abovedave Can't see the changes in the PR — you might need to kick or something to regenerate a PR-env

abovedave:
@CGuindon try now!

CGuindon (approved):
Glorious! Would still wait for Dan before merging though

giamir:
Before merging we should rollback the `package-lock.json` changes and run the formatter to get the lint step to pass. 🙂

Also are we ok to add these bespoke styles here? Should we solve this at Stacks Classic level? We try to keep bespoke editor styles to a minimum in general.

abovedave:
@giamir reverted package changes. They were from some high vulnerability errors on install, but I assume they are the other PR's from dependabot. 

```
12 vulnerabilities (7 moderate, 5 high)
```

If I can get access to this repo I can deal with them if you like?

giamir:
> @giamir reverted package changes. They were from some high vulnerability errors on install, but I assume they are the other PR's from dependabot.

@abovedave I gave you write access to the repo. Feel free to take care of the violations if you have time in a separate PR or commit. Thank you.

giamir:
> Also are we ok to add these bespoke styles here? Should we solve this at Stacks Classic level? We try to keep bespoke editor styles to a minimum in general.

I wanted to wait for Dan's opinion before merging. I personally do think it is something we should solve and document in Stacks Classic: https://github.com/StackExchange/Stacks/blob/develop/lib/components/button-group/button-group.less

@abovedave what are your thoughts? If there is urgency to fix this we can merge this PR and add a follow up ticket to move the styles back in Stacks classic if we agree. If there is no urgency maybe we should directly open a PR in Stacks Classic instead.

abovedave:
The complication is this is a bit of a mash-up component. We're using the radio inputs with the `+` next sibling selector which could have been a [switch component](https://stackoverflow.design/product/components/toggle-switch/#3-or-more-options). So perhaps the 'right' way is to refactor this as a switch, but then add a variation for 'icons only' we've created here.

Another fix might be to refactor this to be more of a pure button group (no radio inputs) and use Stack's atomic classes to achieve the desired visuals. It looks like the outer `:focus` approach we're taking here has been applies to the single switch but not the multiple option.

I think on balance a few lines of CSS here for the high-is priority accessibility fix are probably worth the trade-off vs unpicking the above!

src/styles/custom-components.css
```diff
@@ -59,6 +59,17 @@
     color: var(--black-600) !important;
 }
 
+.s-btn-group .s-btn--radio:checked + .s-btn {
```
giamir, line 62: To keep things extra safe can we scope this rule by targeting `.s-btn-group .s-btn--radio:checked + .s-editor-btn` instead? Sane for the rule below

giamir (commented):
@abovedave I am ok to merge this then, thanks for the explanation. I only have a small ask to ensure the rules are scoped only to the editor. I have added a comment inline. 
After that, we can merge. Thanks.

PR #353, opened 2024-10-01
refactor: Remove liquid templates

abovedave:
I noticed when working on #352 that the Liquid decency is outdated and flagged in a security issue. This PR refactors the preview site templates to use a simpler approach as we're not using any of the features Liquid provides.

### DIFF
--- a/config/webpack.dev.js
+++ b/config/webpack.dev.js
@@ -47,7 +47,7 @@ module.exports = (env, argv) => {
                         templateContent: new Function(
                             ["content"],
                             `return \`${htmlLayout}\`;`
-                        )(String(htmlView)),
+                        )(htmlView),
                         filename: f,
                     });
                 }),
--- a/config/webpack.dev.js
+++ b/config/webpack.dev.js
@@ -3,6 +3,8 @@ const common = require("./webpack.common.js");
 const HtmlWebpackPlugin = require("html-webpack-plugin");
 const fs = require("fs");
 
+const htmlLayout = fs.readFileSync("./site/layout.html");
+
 module.exports = (env, argv) => {
     // add --mode=production to flip this into a pseudo-production server
     const emulateProdServer = argv.mode === "production";
@@ -17,22 +19,6 @@ module.exports = (env, argv) => {
             },
             mode: emulateProdServer ? "production" : "development",
             devtool: emulateProdServer ? false : "inline-source-map",
-            module: {
-                rules: [
-                    {
-                        test: /\.html$/,
-                        use: [
-                            "html-loader",
-                            {
-                                loader: "liquidjs-loader",
-                                options: {
-                                    root: "./site/",
-                                },
-                            },
-                        ],
-                    },
-                ],
-            },
             devServer: {
                 open: false,
                 host:
@@ -51,13 +37,20 @@ module.exports = (env, argv) => {
             },
             plugins: [
                 // create an html page for every item in ./site/views
-                ...fs.readdirSync("./site/views").map(
-                    (f) =>
-                        new HtmlWebpackPlugin({
-                            template: "./site/views/" + f,
-                            filename: f,
-                        })
-                ),
+                ...fs.readdirSync("./site/views").map((f) => {
+                    const htmlView = fs.readFileSync(
+                        `./site/views/${f}`,
+                        "utf8"
+                    );
+
+                    return new HtmlWebpackPlugin({
+                        templateContent: new Function(
+                            ["content"],
+                            `return \`${htmlLayout}\`;`
+                        )(String(htmlView)),
+                        filename: f,
+                    });
+                }),
             ],
             optimization: {
                 splitChunks: {
--- a/package.json
+++ b/package.json
@@ -67,7 +67,6 @@
         "husky": "^9.0.11",
         "jest": "^29.7.0",
         "jest-environment-jsdom": "^29.7.0",
-        "liquidjs-loader": "^1.0.1",
         "mini-css-extract-plugin": "^2.8.1",
         "mini-svg-data-uri": "^1.4.4",
         "postcss": "^8.4.38",
--- a/site/layout.html
+++ b/site/layout.html
@@ -127,11 +127,7 @@
             <label id="a11y-editor-label" class="v-visible-sr">
                 Stacks-Editor demo implementation
             </label>
-            {% block content %}
-            <h1 class="mt32">Minimum viable example</h1>
-            <textarea id="content" name="content" class="d-none"></textarea>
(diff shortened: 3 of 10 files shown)


giamir:
Thanks @abovedave for the PR. I am all up for simplifying but consider that [we have recently introduced liquid templates](https://github.com/StackEng/search-presentation-service/pull/24) for mocking Core's layout in new Free Refills services. I don't see our usage of liquid going away any time soon. 

Perhaps there is another webpack loader we can use instead of removing the liquid templates all together?
Where can I read more about the vulnerabilities of that loader?

Thank you

abovedave:
Ah I see! I could just swap out the `new Function()` with the actual Liquid package and skip the unmaintained loader if we really want it? It just seemed uncessary to use liquid just for a simple layout-view setup.

This was the error:

```
# npm audit report

liquidjs  <10.0.0
Severity: moderate
liquidjs may leak properties of a prototype - https://github.com/advisories/GHSA-45rm-2893-5f49
No fix available
node_modules/liquidjs
  liquidjs-loader  *
  Depends on vulnerable versions of liquidjs
  node_modules/liquidjs-loader

```

giamir:
> It just seemed uncessary to use liquid just for a simple layout-view set 

@abovedave I would like to talk to Dan before removing Liquid all together from this repo. If the `new Function()` approach can buy us some time and fix the vulnerability I would prefer it. Thanks 🙏

b-kelly:
I haven't reviewed this PR yet (expect a review later today), but I'm 100% ok with dropping liquid in this repo. The reason I introduced it was because I wanted a simple shared layout for the docs and liquid syntax was already used in Stacks's docs. The use of liquid in this repo is probably overkill, but it worked ok and stayed out of the way. If it is causing issues now, then let's ditch it.

config/webpack.dev.js
```diff
@@ -3,6 +3,8 @@ const common = require("./webpack.common.js");
 const HtmlWebpackPlugin = require("html-webpack-plugin");
 const fs = require("fs");
 
+const htmlLayout = fs.readFileSync("./site/layout.html");
```
b-kelly, line 6: optional: Move this into the exported function (after line 8) so it only runs when invoked.

In this circumstance it changes basically nothing, but its generally good form to avoid side effects on import when possible.

package.json
```diff
@@ -67,7 +67,6 @@
         "husky": "^9.0.11",
         "jest": "^29.7.0",
         "jest-environment-jsdom": "^29.7.0",
-        "liquidjs-loader": "^1.0.1",
```
b-kelly, line 70: 🎉

site/layout.html
```diff
@@ -127,11 +127,7 @@
             <label id="a11y-editor-label" class="v-visible-sr">
                 Stacks-Editor demo implementation
             </label>
-            {% block content %}
```
b-kelly, line 130: I'll miss having this default example for reference, but... not so much that I'd block this change from going in.

lgtm

config/webpack.dev.js
```diff
@@ -51,13 +37,20 @@ module.exports = (env, argv) => {
...
-                        })
-                ),
+                ...fs.readdirSync("./site/views").map((f) => {
+                    const htmlView = fs.readFileSync(
+                        `./site/views/${f}`,
+                        "utf8"
+                    );
+
+                    return new HtmlWebpackPlugin({
+                        templateContent: new Function(
```
b-kelly, line 47: The use of `new Function` here is a large code smell to me. I'll pull down the PR here shortly and see if I can't whip up an alternative approach.

b-kelly (changes requested):
Other than the use of `new Function`, this change looks good to me. See my [previous comment](https://github.com/StackExchange/Stacks-Editor/pull/353#issuecomment-2391714502) on OK'ing dropping liquid syntax entirely from this repo.

I'll pull this PR down today and find a good alternative to the `new Function` call.

config/webpack.dev.js
```diff
@@ -17,22 +19,6 @@ module.exports = (env, argv) => {
             },
             mode: emulateProdServer ? "production" : "development",
             devtool: emulateProdServer ? false : "inline-source-map",
-            module: {
```
b-kelly, line 20: _loader-update_: Partially revert this change, modifying `html-loader` to do the work for us (GitHub doesn't allow suggestions on deleted lines, so you'll have to fix this manually)

```js
            module: {
                rules: [
                    {
                        test: /\.html$/,
                        use: [
                            {
                                loader: "html-loader",
                                options: {
                                    // wrap the content of the html file in the layout.html file
                                    preprocessor: (content) =>
                                        htmlLayout.replace(
                                            "<%=content%>",
                                            content
                                        ),
                                },
                            },
                        ],
                    },
                ],
            },
```

config/webpack.dev.js
b-kelly, line 6: _loader-update_: Set the `"utf-8"` flag so the file is read as a `string` (instead of `Buffer`)

```suggestion
const htmlLayout = fs.readFileSync("./site/layout.html", "utf-8");
```

b-kelly (changes requested):
I added a few comments with suggested changes. I'd push directly to the branch, but doing so to a fork is a pain in the rear. The change comments are prefixed with `_loader-update_`. The gist of the changes are as follows:

1. readd [`html-loader`](https://webpack.js.org/loaders/html-loader/), since it handles recompiling on file change for us (which [`templateContent` does not](https://github.com/jantimon/html-webpack-plugin#writing-your-own-templates), along with a host of other potentially useful utilities (such as preprocessing, see next point)
2. add a `preprocessor` entry to wrap the files with our layout (inspired by the [Templating section in the loader docs](https://webpack.js.org/loaders/html-loader/#templating))

giamir:
@b-kelly If we are going to keep a templating language of sort, I would prefer to stay consistent with what we use in Stacks docs. We have just started to use liquid also for some free refills mock layout stuff (see [WIP](https://github.com/StackEng/search-presentation-service/tree/main/src/SearchPresentationService.ClientApp/%40stackeng/mock-layout-kit)). Perhaps it is just a matter of swapping the existing loader with one that does not have vulnerabilities. I haven't looked in details but maybe [this one](https://www.npmjs.com/package/liquid-loader) could be an option.

b-kelly:
@giamir I'm ok with adding back in a special loader or templating language, but I think it's overkill for our use case. 100% of the code that does the "templating" in this case is just a single line:

```js
htmlLayout.replace("<%=content%>", content)
```

So... we _really_ don't need a full blown templating solution for this. Imo, we can revisit in the future if we ever need a more full-featured solution.

PR #365, opened 2025-01-14
chore(deps): update dependencies and tidy up typescript typing

threefjefff:
Bumping version of prosemirror and markdown-it to latest. 

This necessitated a cleanup of how we call markdown-it's types and utility functions; rather than reaching into the internals, this now references the public interface.


src/shared/markdown-it/stack-language-comments.ts
```diff
@@ -1,6 +1,5 @@
-import MarkdownIt from "markdown-it/lib";
-import StateBlock from "markdown-it/lib/rules_block/state_block";
-import State from "markdown-it/lib/rules_core/state_core";
+import MarkdownIt from "markdown-it";
+import { StateCore, StateBlock } from "markdown-it";
```
giamir, line 2: We might want to inline this import statement for consistency with the other files.

giamir (changes requested):
Thanks for the PR. ❤️
Changes look reasonable to me so far, but there are some unit and e2e tests failing we should look into before merging. It might also be the upgraded version of the libraries break our test expectations. Hopefully that's not the case because it could make the upgrade not as straightforward. 

Once tests are passing I will take another look. 
Thank you

giamir (approved):
Looks good to me. I will let @dancormier also take a look. If he cannot find anything as well we are good to merge. Thanks a lot for helping us upgrade few dependencies. ❤️

dancormier (approved):
Ran the tests, spun up the editor locally, and checked out the code. Everything ran/looked good to me, so this gets my stamp of approval. Thanks for this @threefjefff!

Issue #234, opened 2022-11-30
Escaping '#' inside bullet lists doesn't round-trip


dionrhys:
**Describe the bug**
Within list items, `#` at the start of the text needs to be escaped to not turn the text into a heading. We can escape that by writing `\#`, but flipping between rich text and markdown removes the backslash, making the text within the list item a heading again.

**To Reproduce**

1. Go to https://editor.stackoverflow.design/
2. Switch to Markdown mode and paste:
   ```markdown
   \# test
   
   - \# test
   
   \# test
   ```
3. Flip to rich text mode, see that `#` is shown as expected
   <details><summary>Screenshot</summary>

   ![Screenshot showing rich text mode the first time](https://user-images.githubusercontent.com/844245/204790652-b7a75c11-3489-4792-a0db-d1928571576d.png)

   </details>

4. Flip to Markdown mode, see that `#` is no longer escaped
   <details><summary>Screenshot</summary>

   ![Screenshot showing markdown mode with changed content](https://user-images.githubusercontent.com/844245/204790682-9fd43bd4-bdb0-4539-aa1c-b4cf3cda8ec5.png)

   </details>

5. Flip to rich text mode, see that the text is now a `<h1>` heading
   <details><summary>Screenshot</summary>

   ![Screenshot showing rich text mode the second time](https://user-images.githubusercontent.com/844245/204790736-18d9e63d-8321-4ce5-b9d6-74134194b1c6.png)

   </details>

**Expected behavior**
The editor should keep the needed backslash escape on the `#` character so that it doesn't turn the text into a heading.

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Firefox 
 - Version: 107.0

giamir:
Resolved by https://github.com/StackExchange/Stacks-Editor/pull/366

PR #366, opened 2025-01-16
Escape roundtrip

threefjefff:
Closes [#234]

Treat the start of inlining a parent block the same way as the begining of a line (to catch sequences where, for example, we've fought with the editor in rich text mode to allow us to insert a `#` without it turning into a header, thereby serialising to `- \#`)

**PR Checklist**

**Environment(s) tested**

- Device: descktop
- OS: Windows
- Browser: Chrome
- Version 132.0.6834.160 (Official Build) (64-bit) (cohort: Stable)


giamir (commented):
Thanks @threefjefff for the PR. 
Overall I think the PR is going in the right direction. 🎉

You should add a unit test for the `preserve-escape` rule and see if it is worth adding a regression test for the issue Dion reported. Perhaps in the markdown-serializer unit test file. 

Great work so far! You can see how much code you ended up writing to fix a relatively simple bug - this is why we should be intentional going forward about what is worth fixing and what is not. More code = more maintenance, more cognitive overhead, more bugs potential, etc...

For this specific bug I think we can go ahead and merge once the PR is in a good state with its relevant tests. Thanks for now. 🙂

dancormier (commented):
Thanks for the bug fix @threefjefff. I ran `npm run format` to resolve some of the formatting inconsistencies and get the [lint job](https://github.com/StackExchange/Stacks-Editor/actions/runs/12837218294/job/35800436897?pr=366) passing (https://github.com/StackExchange/Stacks-Editor/pull/366/commits/8c7a3e02b034596fe20bb685c2236ae15728c152). Other than that, the only things I think necessary to get this merged is what Giamir mentioned.

test/shared/roundtrip.e2e.test.ts
```diff
@@ -0,0 +1,56 @@
...
+        ["lol\n\n***\n\nlmao"],
+        ["`code`"],
+        //TODO: Codeblock does weird things roundtripping: Adds an extra space
+        //['```javascript\ncodeblock\n```' ],
+
+        //Escape character
+        [String.raw`\# not a header`],
+        [String.raw`- \# list item (not header)`],
+    ] as const) {
+        test(`should make markdown -> richtext -> markdown round trip '${markdown}'`, async () => {
```
giamir, line 47: This test title is making the test output less readable because it honours the `\n` special char. A way to solve that is to do something like this:
```
`should make markdown -> richtext -> markdown round trip ${JSON.stringify(markdown)}`
```
Screenshot 2025-02-04 at 10 15 34

test/shared/roundtrip.e2e.test.ts
```diff
@@ -0,0 +1,56 @@
...
+        ["> blockquote"],
+        ["* List Item"],
+        ["- List Item"],
+        ["1. List Item"],
+        ["2) List Item"],
+        ["lol\n\n---\n\nlmao"],
+        ["lol\n\n***\n\nlmao"],
+        ["`code`"],
+        //TODO: Codeblock does weird things roundtripping: Adds an extra space
+        //['```javascript\ncodeblock\n```' ],
```
giamir, line 41: 👍 Good knowledge for a potential follow up ticket (in case it is worth fixing of course).

giamir (changes requested):
Thanks @threefjefff for adding the tests.

As I was reviewing the tests I realized that original issue reported seems to be fixed exclusively by the `state.atBlockStart` expression in the markdown serializer you have added:
```ts
const startOfLine: boolean = state.atBlank() || state.atBlockStart || state.closed;
```

If that's the case we could reduce significantly the amount of code we will need to merge in. We could just add that expression and keep the `roundtrip.e2e.test` you created and call it done. 

I don't know if I am missing some edge cases;  as far as I can tell the tests seem to confirm my theory of the bug being fixed by simply adding `state.atBlockStart`. 🤷

Also sorry for realizing this only now.

threefjefff:
> As I was reviewing the tests I realized that original issue reported seems to be fixed exclusively by the `state.atBlockStart` expression in the markdown serializer you have added:

Great catch! I think this is one of those things where I started fixing things and then was blinded by the implementation. Such is the benefit of a second pair of eyes 🤦‍♂️

> I don't know if I am missing some edge cases; as far as I can tell the tests seem to confirm my theory of the bug being fixed by simply adding `state.atBlockStart`. 🤷

Certainly not for this case. We were discussing these "feel bad" roundtrip errors the other day, and it might be that what's here is a precursor for some of those, but if so I can always refer back to earlier states of this PR. 
Much further down the line, we might want to also styalize and add hover behaviour for comments, but that's very much not a priority at the moment.

I'll get to stripping this back to the essentials and get back to you.

giamir (approved):
Great stuff @threefjefff. 
I am happy we could sort this issue with less than a line of code in the end. 🎉

I am merging this in.

Issue #64, opened 2021-03-17
Insert a space after certain marks on left arrow key press so they don't trap the cursor.


b-kelly:
This is a meta-entry to track items such as #31 and #49.

**Is your feature request related to a problem? Please describe.**

Specific marks (e.g. code_inline and kbd) appear to "trap" the cursor, even though their behavior is consistent to other marks (e.g. italics or bold). The common behavior is that activating a mark (e.g. ctrl+i for italics) will keep the cursor in that typing mode until it is again deactivated. You can see this behavior in other editors as well. However, some marks that seem like standalone chunks (kbd, code_inline) trigger an expectation in users that they can "leave" these chunks at will by manually moving the cursor with the arrow keys. You can see this behavior in Slack's `code` formatting.

**Describe the solution you'd like**

For a small subset of marks, when the user uses the right arrow key to move outside of it _and there is no other content to the right of it_, disable the mark and add a `space`. This is consistent with the behavior seen in Slack.

**Describe alternatives you've considered**

We could just _not_ do anything and consider all marks to behave the same. However, since we have received independent user reports about this behavior, this would likely not be the most UX friendly option.

PR #383, opened 2025-02-26
fix(code-blocks): remove buggy "exit code block" functionality

### LINKED ISSUE #191: bug(rich-text) down-arrow in code block exits the code block even when more lines in the code block below cursor
**Describe the bug**
Pressing the down arrow key in a code blocks with multiple lines moves out of the code block, instead of down to the next line in the code block.

**To Reproduce**
1. RT mode
2. Add text with multiple lines, make into code block
3. Cursor on the top line
4. Down arrow on keyboard
5. Observe: cursor skips remaining lines in the code block, exits code block

**Expected behavior**
If additional lines below cursor in the code block, down arrow should move to next line. Down arrow should only exit code block when cursor is on the last line in the code block.

**Screenshots**
demo

In demo, user pressed: down up up

**Smartphone (please complete the following information):**

**Additional context**

alexwarren:
Closes #191

**Describe your changes**

This PR removes the custom logic that was attached to "arrow right" and "arrow down" keypresses.

These were added in #64 in an attempt to add some behaviour where people could exit code blocks by pressing arrow keys. Unfortunately that introduces bugs like this one, where you can't use the arrow keys to navigate _inside_ a code block any more - instead you get teleported out of it. Annoying!

After removing this code, things seem to work as I would expect them to anyway.

https://github.com/user-attachments/assets/a3186241-4d82-4353-8a30-cd4439462207

**PR Checklist**

**Environment(s) tested**

Tested on Chrome and Safari.

**Additional context**

n/a

### DIFF
--- a/test/rich-text/commands/index.test.ts
+++ b/test/rich-text/commands/index.test.ts
@@ -785,12 +785,20 @@ describe("commands", () => {
         );
 
         it("should handle the case when $cursor is null", () => {
+            // suppress console.warn
+            const consoleWarnSpy = jest
+                .spyOn(console, "warn")
+                .mockImplementation(() => {});
+
             let state = createState("this is my state", []);
             state = state.apply(
                 state.tr.setSelection(TextSelection.create(state.doc, 0, null))
             );
             expect((<TextSelection>state.selection).$cursor).toBeNull();
             expect(() => exitInclusiveMarkCommand(state, null)).not.toThrow();
+
+            // restore console.warn
+            consoleWarnSpy.mockRestore();
         });
     });
 
--- a/test/rich-text/commands/index.test.ts
+++ b/test/rich-text/commands/index.test.ts
@@ -1,4 +1,4 @@
-import { EditorState, Transaction } from "prosemirror-state";
+import { EditorState, TextSelection, Transaction } from "prosemirror-state";
 import {
     exitInclusiveMarkCommand,
     insertRichTextHorizontalRuleCommand,
@@ -733,7 +733,7 @@ describe("commands", () => {
         );
     });
 
-    describe("exitMarkCommand", () => {
+    describe("exitInclusiveMarkCommand", () => {
         it("all exitable marks should also be inclusive: true", () => {
             Object.keys(testRichTextSchema.marks).forEach((markName) => {
                 const mark = testRichTextSchema.marks[markName];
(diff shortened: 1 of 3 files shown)


alexwarren:
@dancormier Thanks, I've reinstated that.

I've added a test for the case when `$cursor` is null - I've sort of hacked my way around to get there though, and the test generates a console warning "TextSelection endpoint not pointing into a node with inline content (doc)". Which I guess is intentional, I'm deliberately creating an invalid state here (?) to ensure our code can handle it. But I'm not sure if there's a better way of testing this.

test/rich-text/commands/index.test.ts
```diff
@@ -783,6 +783,15 @@ describe("commands", () => {
...
         );
+
+        it("should handle the case when $cursor is null", () => {
+            let state = createState("this is my state", []);
+            state = state.apply(
+                state.tr.setSelection(TextSelection.create(state.doc, 0, null))
+            );
+            expect((<TextSelection>state.selection).$cursor).toBeNull();
+            expect(() => exitInclusiveMarkCommand(state, null)).not.toThrow();
+        });
```
dancormier, line 794: > I've added a test for the case when $cursor is null - I've sort of hacked my way around to get there though, and the test generates a console warning "TextSelection endpoint not pointing into a node with inline content (doc)". Which I guess is intentional, I'm deliberately creating an invalid state here (?) to ensure our code can handle it. But I'm not sure if there's a better way of testing this.

You can suppress `console.warn` with `jest.spyOn` like this:

```suggestion
        it("should handle the case when $cursor is null", () => {
            // suppress console.warn
            const consoleWarnSpy = jest
                .spyOn(console, "warn")
                .mockImplementation(() => {});

            let state = createState("this is my state", []);
            state = state.apply(
                state.tr.setSelection(TextSelection.create(state.doc, 0, null))
            );
            expect((<TextSelection>state.selection).$cursor).toBeNull();
            expect(() => exitInclusiveMarkCommand(state, null)).not.toThrow();

            // restore console.warn
            consoleWarnSpy.mockRestore();
        });
```

No clue why it won't show the diff the way I'd expect, but the lines added are at the start and end of the test only.

dancormier (changes requested):
Thanks for making those changes @alexwarren. I've suggested one minor change to get rid of that console.warn statement that comes up when you run the test. Otherwise, it looks good and I'd be happy to see it merged once the one thing is resolved

alexwarren:
Thanks @dancormier, I've made that change.

dancormier (approved):

PR #384, opened 2025-02-28
fix(code-blocks): press Enter at start of first code block to create paragraph

### LINKED ISSUE #197: bug(rich-text) no way to advance above a code block that is the first element
**Describe the bug**
This bug is the opposite of #41. In rich text mode, if the first element in a post is a source fence, there's no way to insert a blank line before it, anything you type gets inserted inside the fence. (The proposed Shift + Enter workaround escapes the fence downward but not upward.)

**To Reproduce**
1. RT mode, write some text, make it into a code block
2. Attempt to advance to _before_ the code block. Fail at doing so.

**Expected behavior**
There is some way to advance the cursor to be before the code block, so as to add text or some other block element. The only way to do it now is in MD mode.

**Screenshots**
demo

Windows 11 Chrome 103

**Additional context**
https://meta.stackexchange.com/a/380418/51

alexwarren:
Closes #197

**Describe your changes**

This PR make it so that if a code block is the first thing in a document, then when you press Enter at the start of that code block, a paragraph is created above.

This _only_ applies if you're at the start of the code block, _and_ only if that code block is the first thing in the doc.

https://github.com/user-attachments/assets/5116c278-ce1f-41e4-816b-7002198c1745

**PR Checklist**

**Environment(s) tested**

Chrome, Safari

### DIFF
--- a/src/rich-text/commands/index.ts
+++ b/src/rich-text/commands/index.ts
@@ -582,11 +582,10 @@ export function exitInclusiveMarkCommand(
     return true;
 }
 
-export function maybeSplitCodeBlock(state: EditorState, dispatch: (tr: Transaction) => void) {
-    console.log('maybeSplitCodeBlock');
+export function splitCodeBlockAtStart(state: EditorState, dispatch: (tr: Transaction) => void) {
     const { $from } = state.selection;
-    if ($from.parentOffset === 0) {
+    if ($from.parent.type.name === "code_block" && $from.parentOffset === 0) {
         return splitBlock(state, dispatch);
     }
     return false;
-  }
+}
--- a/src/rich-text/commands/index.ts
+++ b/src/rich-text/commands/index.ts
@@ -1,4 +1,4 @@
-import { setBlockType, toggleMark, wrapIn } from "prosemirror-commands";
+import { setBlockType, splitBlock, toggleMark, wrapIn } from "prosemirror-commands";
 import { Mark, MarkType, NodeType, Schema } from "prosemirror-model";
 import {
     Command,
@@ -581,3 +581,12 @@ export function exitInclusiveMarkCommand(
 
     return true;
 }
+
+export function maybeSplitCodeBlock(state: EditorState, dispatch: (tr: Transaction) => void) {
+    console.log('maybeSplitCodeBlock');
+    const { $from } = state.selection;
+    if ($from.parentOffset === 0) {
+        return splitBlock(state, dispatch);
+    }
+    return false;
+  }
--- a/src/rich-text/commands/index.ts
+++ b/src/rich-text/commands/index.ts
@@ -1,4 +1,9 @@
-import { setBlockType, splitBlock, toggleMark, wrapIn } from "prosemirror-commands";
+import {
+    setBlockType,
+    splitBlock,
+    toggleMark,
+    wrapIn,
+} from "prosemirror-commands";
 import { Mark, MarkType, NodeType, Schema } from "prosemirror-model";
 import {
     Command,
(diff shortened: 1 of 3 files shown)


src/rich-text/commands/index.ts
```diff
@@ -581,3 +586,26 @@ export function exitInclusiveMarkCommand(
...
+    if ($from.parentOffset !== 0) {
+        return false;
+    }
+
+    // Is this code block the first child of the doc (i.e. no other nodes above it)?
+    if ($from.depth !== 1 || $from.index(0) !== 0) {
+        return false;
+    }
+
+    return splitBlock(state, dispatch);
```
dancormier, line 610: As a user, I'd expect the cursor to move into the newly created block above the code block. It's slight but I think this will provide a smoother experience.

Here's the code that moves the cursor to the start of the doc:

```suggestion
    if (dispatch) {
        let tr = state.tr.split($from.pos);
        tr = tr.setNodeMarkup($from.before(), state.schema.nodes.paragraph);
        tr = tr.setSelection(TextSelection.create(tr.doc, $from.pos - 1));
        dispatch(tr);
    }

    return true;
```

If there's an argument against this, let me know!

test/rich-text/commands/index.test.ts
```diff
@@ -1238,4 +1239,92 @@ describe("commands", () => {
...
+            expect(newState.doc.childCount).toBe(2);
+
+            // The first node should be an empty paragraph:
+            expect(newState.doc.firstChild.type.name).toBe("paragraph");
+            expect(newState.doc.firstChild.textContent).toBe("");
+
+            // The second node should be the code block with the original text:
+            expect(newState.doc.lastChild.type.name).toBe("code_block");
+            expect(newState.doc.lastChild.textContent).toBe("Some code");
+        });
```
dancormier, line 1269: Here's a little test modification to support the change I suggested earlier in this review:

```suggestion

            // The cursor should be at the start of the newly created paragraph.
            expect(newState.selection.from).toBe(0);
            expect(newState.selection.to).toBe(0);
        });
```

dancormier (changes requested):
Nice work @alexwarren! I looked through the code seems solid. Only once minor suggested change on my part. Once that's resolved, I'm good to see this merged. Thanks!

### Out of scope

I start to get a little scared when I see files with more than a couple hundred lines. Long term, I think we'd be wise to separate `src/rich-text/commands/index.ts` and the related test file into per-function files. This would probably make maintenance and modification easier.

src/rich-text/commands/index.ts
giamir: @dancormier Personally I am not sure about this. I would probably keep it as is for now since it's also less code and we can add this extra behavior later if necessary.

Moving the cursor up when you press enter it's not what most WYSIWYG editors seem to do. For example when you are at the beginning of the first item in a bullet list in google docs and you press enter, then the cursor will create a new bullet above and cursor will move down; it is up to the user to move the cursor back up and remove the bullet with the delete key. 

In fact even the solution as is now is a bit debatable for me in terms of expected experience. What if I need to add a line of code at the top of the code block I created? Now I will be pushed out of the code block when I press enter (bad, but I can work around it at least compared to the previous experience where it was not possible at all to exit the code block). A potentially more intuitive approach could be to treat the codeblock as the jira editor does (like an image essentially, you can exit that context with the arrow keys) but  that will likely be a bit more expensive to implement. I am happy to merge this as is as a stop gap but this is where input from a designer that have done some comparisons with other editors could help.

dancormier (approved):

PR #389, opened 2025-03-04
fix(preview): Fix constant layout shift as preview rerenders

alizaberger:
**Describe your changes**

Fixes the issue described in [this meta post](https://meta.stackoverflow.com/questions/431137/stacks-editor-is-doing-rapid-layout-shift-as-i-edit-the-body-input). As the user types, preview container content is removed before the promise is resolved, causing the page layout to shift during the time it takes for the promise to resolve. The fix is to not clear the preview, rather wait for the content to be replaced when the promise is resolved. 

**Before**

[screen-capture (1).webm](https://github.com/user-attachments/assets/de33cf85-6be9-4d0b-a58a-3d389dfef417)

**After**

[screen-capture.webm](https://github.com/user-attachments/assets/74cc0864-59f8-4dfe-8a92-935b14587d78)

**PR Checklist**

**Environment(s) tested**

- Device: [e.g. desktop, iPhone6, etc.]

**Additional context**

Add any other context about the PR here.

### DIFF
--- a/site/index.ts
+++ b/site/index.ts
@@ -121,12 +121,6 @@ const ImageUploadHandler: ImageUploadOptions["handler"] = (file) =>
  * NOTE: synchronous renderers can simply return Promise.resolve()
  */
 const examplePreviewRenderer: PreviewRenderer = async (content, container) => {
-    const spinner = document.createElement("span");
-    spinner.className = "is-loading";
-    spinner.textContent =
-        "Intentionally delaying render for example purposes...";
-    container.appendChild(spinner);
-
     // add a fake load delay (because we can)
     await sleepAsync(500);
 
--- a/site/index.ts
+++ b/site/index.ts
@@ -120,29 +120,15 @@ const ImageUploadHandler: ImageUploadOptions["handler"] = (file) =>
  * Sample preview renderer that has a fake delay and uses the default Markdown-It renderer
  * NOTE: synchronous renderers can simply return Promise.resolve()
  */
-const examplePreviewRenderer: PreviewRenderer = async (content, container, abortSignal) => {
+const examplePreviewRenderer: PreviewRenderer = async (content, container) => {
     const spinner = document.createElement("span");
     spinner.className = "is-loading";
     spinner.textContent =
         "Intentionally delaying render for example purposes...";
     container.appendChild(spinner);
 
-    const interruptableSleepAsync = (delayMs: number, signal: AbortSignal) => new Promise<void>((resolve, reject) => {
-        const timeoutId = setTimeout(() => {
-            resolve();
-        }, delayMs);
-
-        signal.addEventListener("abort", () => {
-            clearTimeout(timeoutId);
-            reject(new Error('aborted'));
-        });
-    });
-
-    await interruptableSleepAsync(500, abortSignal)
-        .catch(() => {
-            container.removeChild(spinner);
-        })
-
+    // add a fake load delay (because we can)
+    await sleepAsync(500);
 
     const instance = MarkdownIt("commonmark", {
         html: false,
--- a/site/index.ts
+++ b/site/index.ts
@@ -120,15 +120,29 @@ const ImageUploadHandler: ImageUploadOptions["handler"] = (file) =>
  * Sample preview renderer that has a fake delay and uses the default Markdown-It renderer
  * NOTE: synchronous renderers can simply return Promise.resolve()
  */
-const examplePreviewRenderer: PreviewRenderer = async (content, container) => {
+const examplePreviewRenderer: PreviewRenderer = async (content, container, abortSignal) => {
     const spinner = document.createElement("span");
     spinner.className = "is-loading";
     spinner.textContent =
         "Intentionally delaying render for example purposes...";
     container.appendChild(spinner);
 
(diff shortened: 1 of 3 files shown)


alexwarren:
One regression I notice with this change is that we can end up with multiple spinners created by `examplePreviewRenderer`:

https://github.com/user-attachments/assets/ae6195cb-9592-4e46-88fc-4010bdc5743d

Maybe that was why there was this code for clearing out the preview in the first place?

I'm not sure if that means there is something to be fixed in `examplePreviewRenderer` itself, or if there needs to be some way of cancelling a call to the `preview.renderer` if there is one in-flight when a newer edit is made.

giamir:
> I'm not sure if that means there is something to be fixed in examplePreviewRenderer itself, or if there needs to be some way of cancelling a call to the preview.renderer if there is one in-flight when a newer edit is made.

Renderers are async functions: in the current implementation once we call them there is no way to cancel the side effects that are tasked to do from the editor codebase unfortunately. What we can do though is to pass in an [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) signal as option that the consumers implementing the renderer can leverage to abort whatever they are doing.

This is how the update preview method will change (notice the addition of a new abortController class member)
```js
private updatePreview(text: string) {
    this.container.innerHTML = "";

    // if showing the preview, fire off the renderer async
    if (this.isShown) {
        
       // PART OF CODE REMOVED BY THIS PR
           
        this.container.appendChild(this.dom);

        // Abort any previous render
        if (this.abortController) {
            this.abortController.abort();
        }

        // Create a new AbortController for the current render
        this.abortController = new AbortController();
        const signal = this.abortController.signal;

        void this.renderer?.(text, this.dom, { signal }).catch((e) => {
            if (e.name !== 'AbortError') {
                error(
                    "PreviewView.updatePreview",
                    `Uncaught exception in preview renderer`,
                    e
                );
            }
        });
    }
}
```

Then on the consumer end you can consume the signal. In the case of our `examplePreviewRenderer` in the site/index.ts file the part of `sleepAsync(500)` would become:

```js
const interruptableSleepAsync = (delayMs, signal) => new Promise((resolve, reject) => {
        const timeoutId = setTimeout(() => {
            resolve();
        },  delayMs);

        signal.addEventListener("abort", () => {
            clearTimeout(timeoutId);
             reject(new Error('aborted'));
         });
 });
 
 interruptableSleepAsync(500, signal)
 .then(() => {
   // do what we were doing before
 })
 .catch(() => {
   // cleanup any lingering dom elements added previously if necessary (e.g. remove the spinner)
 })
```

Now you need to consider that in Core the renderer is different and a bit more sophisticated / legacy.
https://github.com/StackEng/StackOverflow/blob/main/StackOverflow/_Scripts/PartialJS/stacks-editor.mod.ts#L514-L520
Code will need to be adjusted in there so that the renderer can accept the signal and use it to prevent async side effects from happening when an in-flight render is aborted.

I am tagging @b-kelly since I don't know why we don't have already a mechanism to instruct consumers preview renderers to interrupt in-flight rendering. Perhaps we were relying on the fact that race conditions for async renders fire one after the other would be unlikely.

alizaberger:
@giamir That looks like a nice pattern to interrupt the renderer, I will work on implementing that

alizaberger:
Here's what the preview pane looks like with the AbortController at work @giamir @alexwarren 

[screen-capture (2).webm](https://github.com/user-attachments/assets/cf5deb31-c5f7-4394-a6d7-6283a8119ea0)

giamir:
@alizaberger before investing time on implementing the mechanism to interrupt in-flight rendering it would be interesting to know if that would be useful in our Core context in my opinion. 

You could go in Core to `StackOverflow/node_modules/@stackoverflow/stacks-editor/dist/commonmark/plugins/preview.js` and remove line 71 and 72 as you have done in this PR and then run `npm start` and check the specific page where the bug was reproduced.

giamir:
> Here's what the preview pane looks like with the AbortController at work @giamir @alexwarren
 
I think that's what I would expect based on how the examplePreviewRenderer is implemented in the demo site. We should not based our decision though on what that renderer but rather on what the renderer of Core does. This is why I was suggesting to check how your small changes behave in the Core context. If it does its job and doesn't have strange spinner appearing I would argue to keep the initial changes you made and adjust the implementation of the examplePreviewRenderer of the demo site so that we don't get that spinner cascade (perhaps we could remove the spinner all together from the renderer implementation). Hope my suggestion makes sense. If not feel free to DM me. :)

alizaberger:
Final screen capture:

[screen-capture (3).webm](https://github.com/user-attachments/assets/f5de4931-a0ad-42f7-9fb2-5ecfaea9c690)

dancormier (approved):
Thanks for the PR @alizaberger! I tested this change both in the standalone editor and by duplicating this change the editor in Core. Everything worked as I expected 🎉



My only (optional) suggestion is adding a test for this functionality. I'm fine with seeing this PR merged in as-is, but I think it could be worth trying to add tests to any code that's touched that doesn't already include a test case. Since this change is so minor, I'm ok with it not getting a test at the moment, but if you have a cycle or two to try an add one, I think it would be handy to have added.

PR #393, opened 2025-03-11
fix(markdown): Fix code block selection off by one error

alizaberger:
**Describe your changes**

Fixes the issue described in [this meta post](https://meta.stackoverflow.com/questions/431670/creating-a-code-block-in-staging-ground-includes-one-additional-character-after). In markdown mode, the code block button surrounds one character extra than the user's selection. This PR corrects that, cleans up the code, and adds comments to make it more understandable. 

**PR Checklist**

**Environment(s) tested**

### DIFF
--- a/src/commonmark/commands.ts
+++ b/src/commonmark/commands.ts
@@ -439,8 +439,8 @@ function blockWrapIn(
     to += 1;
 
    //always insert a newline after the selection
-        tr = tr.insertText(surroundingChar, to, to);
-        to += 1;
+    tr = tr.insertText(surroundingChar, to, to);
+    to += 1;
    
     // insert the code fences from the end first so we don't mess up our from index
     //consider if we need to add the surrounding char too, if the text is not already surrounded by newline
@@ -453,7 +453,6 @@ function blockWrapIn(
 
     // set the selection to the text inside the code fences, skipping the leading newline
     if (dispatch) {
-
         tr.setSelection(
             TextSelection.create(
                 state.apply(tr).doc,
--- a/src/commonmark/commands.ts
+++ b/src/commonmark/commands.ts
@@ -445,12 +445,19 @@ function blockWrapIn(
     // insert the code fences from the end first so we don't mess up our from index
     //consider if we need to add the surrounding char too, if the text is not already surrounded by newline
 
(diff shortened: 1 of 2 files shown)


src/commonmark/commands.ts
```diff
@@ -416,57 +416,49 @@ function blockWrapIn(
...
             state,
             dispatch
         );
     }
 
-    let { from, to } = state.selection;
+    const from = state.selection.from;
+    let to = state.selection.to;
 
-    // check if we need to unwrap
```
alizaberger, line 428: Is already done in the calling function

src/commonmark/commands.ts
```diff
@@ -416,57 +416,47 @@ function blockWrapIn(
...
-            surroundingChar.length + formattingText.length;
+    const preceedingNewlineNeeded = from > 0 && state.doc.textBetween(from - 1, from) !== surroundingChar ? surroundingChar : '';
+    const followingNewlineNeeded = to + 1 < state.doc.content.size && state.doc.textBetween(to, to + 1) !== surroundingChar ? surroundingChar : '';
 
+    tr.insertText(formattingText + followingNewlineNeeded, to);
+    tr.insertText(preceedingNewlineNeeded + formattingText, from);
+    to += (formattingText.length * 2) + preceedingNewlineNeeded.length;
+
+    // set the selection to the text inside the code fences, skipping the leading newline
+    if (dispatch) {
```
threefjefff, line 454: While we're editing code here, can we also do a check for is `dispatch` is defined earlier in the code? We can skip a lot of pointless transaction building if we establish early whether or not this action is valid, and that seems possible by the time we've got to `let tr = state.tr;`

From [here](https://prosemirror.net/docs/guide/#commands):
> To be able to query whether a command is applicable for a given state, without actually executing it, the dispatch argument is optional—commands should simply return true without doing anything when they are applicable but no dispatch argument is given

threefjefff (approved):
Code looks good, tests look good, some potential for additional cleanup.

dancormier (changes requested):
I tested and this works well. It's nice to see this bug get fixed while simplifying the code. 

Only minor suggestion is to give `npm run format` which will resolve the CI lint failure. I'll get you an approval asap once the pipeline succeeds.

Other than that, this seems good to go IMO! Thanks for the bug fix

threefjefff (approved):

PR #394, opened 2025-03-12
fix(inline-code): disallow "inline code" when selection has linebreaks

alexwarren:
**Describe your changes**

This PR addresses an issue raised on Meta here: https://meta.stackoverflow.com/questions/430988/what-causes-single-backticks-on-every-line-of-code

To address this, we disable the "inline code" button when the text selection includes linebreaks. Users should use code blocks instead of inline code when there are multiple lines.

The fix applies whether there are "real" linebreaks in the selection, or if there are "soft breaks" - these are what you get when you paste code into the Markdown view, then switch back to the rich text mode.

To test this, paste in the example code block from the Meta post:

```
from modules.prompt import create_conversation_prompt
from langchain.chains import ConversationChain
from langchain_core.output_parsers import StrOutputParser
from langchain_core.runnables import RunnablePassthrough
```

Note - this is _not_ currently detected as a code block by the rich text editor. So that makes it a good test for this.

Previously you would be able to toggle "inline text" on this, and get a bunch of paragraphs with backticks around each one. Now, the "inline code" button is disabled. Use the "code block" button to make it a proper code block instead.

### DIFF
--- a/src/rich-text/commands/code-block.ts
+++ b/src/rich-text/commands/code-block.ts
@@ -1,6 +1,7 @@
 import { Node as ProseMirrorNode, NodeType, Schema } from "prosemirror-model";
 import { EditorState, Transaction, TextSelection } from "prosemirror-state";
 import { insertParagraphIfAtDocEnd, safeSetSelection } from "./helpers";
+import { toggleMark } from "prosemirror-commands";
 
 // indent code with four [SPACE] characters (hope you aren't a "tabs" person)
 const CODE_INDENT_STR = "    ";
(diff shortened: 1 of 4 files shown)


threefjefff (approved):

giamir (approved):

dancormier (approved):
Disallowing inline code when the selected text spans blocks seems like a reasonable approach to me. I tested this in [both](https://meta.stackoverflow.com/a/430990/652353) "Reproduction steps" [answers](https://meta.stackoverflow.com/a/430989/652353) in the meta post and everything worked as I'd expect.

Code changes seem straight forward and the tests make sense. Thanks for this @alexwarren

PR #396, opened 2025-03-14
fix(code-blocks): add empty paragraph at end when using three backticks to create code block

alexwarren:
This PR fixes one of the issues reported on Meta: https://meta.stackexchange.com/questions/407446/new-stacks-editors-code-block-usage-issues

Existing functionality: When pressing the Code Block button in the toolbar, a code block is created, and a new empty paragraph is created underneath, if we're at the end of the document. That means you can press the Down arrow to exit the code block.

The problem: This didn't work when creating a code block by entering three backticks: ```

The fix: Call the same `insertParagraphIfAtDocEnd` function when creating a code block using back ticks. This requires wrapping the `textblockTypeInputRule`.

https://github.com/user-attachments/assets/502e68ac-2534-4ebf-b41d-24a2b3ec313b

### DIFF
--- a/src/rich-text/inputrules.ts
+++ b/src/rich-text/inputrules.ts
@@ -4,9 +4,10 @@ import {
     textblockTypeInputRule,
     wrappingInputRule,
 } from "prosemirror-inputrules";
-import { MarkType, Schema } from "prosemirror-model";
-import { EditorState } from "prosemirror-state";
+import { MarkType, NodeType, Schema } from "prosemirror-model";
+import { EditorState, Transaction } from "prosemirror-state";
 import { CommonmarkParserFeatures } from "../shared/view";
+import { insertParagraphIfAtDocEnd } from "./commands/helpers";
 
 // matches: `some text`, but not ` text `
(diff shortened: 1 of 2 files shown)


src/rich-text/inputrules.ts
```diff
@@ -113,7 +162,7 @@ export const richTextInputRules = (
         (match) => ({ level: match[1].length })
     );
 
-    const codeBlockRule = textblockTypeInputRule(
+    const codeBlockRule = textblockTypeTrailingParagraphInputRule(
```
threefjefff, line 165: This covers the case where you're working in RT mode, but not when parsing from markdown. I think the answer here is to capture the keyboard-event time, rather than block-creation time.

only_rt

threefjefff (changes requested):

src/rich-text/inputrules.ts
alexwarren: That's not the issue reported on Meta though, and I think if you were in Markdown mode, it's more reasonable to expect you to switch back.

I think changing all of this to add paragraphs when moving the cursor around is a much bigger change that is going to involve undoing other recent changes.

PR #400, opened 2025-03-19
feat: Add support for Stack Snippets

threefjefff:
Add Stacks Snippet support to the editor as an external plugin.

- Button to open an external modal, passing existing Stack Snippet if selected
- Added to Prosemirror and Markdown IT schemas, so that snippets are correctly parsed and serialized between the two types of Node/Token streams.
- Added a paste handler to convert pasted snippet code directly into the rich text view
- Added the ability to append new content (expect to be used by the external modal)
- Added a snippet view, which is used to render the blocks of snippet content, and a "Run Code" button to render the content to IFrame
- DEVX: added a route for snippet contents such that the render button uses the webpack proxy to POST to the Stacks Snippets service when using the test harness.

**PR Checklist**

**Environment(s) tested**

- Device: desktop
- OS: Windows
- Browser Chrome
- Version 134.0.6998.37

**Additional context**

This is the baseline for Stack Snippets:
- The buttons will not appear unless a renderer option is supplied (which it is not, currently)
- The button use a placeholder icon, this is going through internal design at the moment
- Continuing to do internal UX review and squashing any bugs found. Intention is to triage and roll forwards with bugs that aren't showstoppers.


package.json
```diff
@@ -104,7 +104,8 @@
         "prosemirror-state": "^1.4.3",
         "prosemirror-transform": "^1.8.0",
         "prosemirror-utils": "^1.2.1-0",
-        "prosemirror-view": "^1.37.1"
+        "prosemirror-view": "^1.37.1",
+        "ts-md5": "^1.3.1"
```
alexwarren, line 108: Not sure why this was added, it doesn't look like you're creating MD5 hashes here?

src/rich-text/key-bindings.ts
```diff
@@ -89,6 +90,7 @@ export function allKeymaps(
         "Mod-'": toggleMark(schema.marks.kbd),
         // exit inline code block using the right arrow key
         "ArrowRight": exitInclusiveMarkCommand,
+        "ArrowDown": escapeUnselectableCommand,
```
alexwarren, line 93: Nice, does this mean we can remove the things that create paragraphs after code blocks? Maybe not in this PR but in a subsequent one.

package.json
threefjefff: Ah, good catch. An earlier version was using it to assert that the code blocks hadn't changed between renders.

src/rich-text/key-bindings.ts
threefjefff: That's the general aim, yeah. I scoped it very specifically around this unselectable container as part of this (see: it's already a beefy lad).

What I'd like to see is us move from creating things with extra paragraphs at time of creation, and move to this "just-in-time" style, because then parsing/serializing doesn't have to worry about extra paragraph nodes for the same of useability; it can focus on the document at hand. This is an experiment into that idea.

threefjefff:
After discussion, we're going to amend the external plugin architecture to make this work as an external plugin. This also means moving the `/plugins` folder to the top level, and we can think about distrubtion strategies at a later date.

src/shared/editor-plugin.ts
```diff
@@ -253,52 +257,36 @@ export class ExternalPluginProvider implements IExternalPluginProvider {
 
     /** {@inheritDoc IExternalPluginProvider.getFinalizedMenu} */
     getFinalizedMenu(menu: MenuBlock[], schema: Schema): MenuBlock[] {
```
threefjefff, line 259: Found that this mutates `menu` when passed in, so rewrote to be idempotent.

src/shared/menu/helpers.ts
```diff
@@ -82,6 +82,14 @@ export interface DropdownMenuItem extends BaseMenuItem {
  */
 export type MenuItem = BaseMenuItem | DropdownMenuItem;
 
+export type CoreMenuBlockNames =
```
threefjefff, line 85: What I'm attempting here is to have known keys for plugin developers to merge with. Not sure if this is strictly necessary, given I ended up going the root of returning an amended version of the menus as a passed-in parameter.

src/shared/localization.ts
```diff
@@ -43,6 +43,10 @@ export const defaultStrings = {
         sub: shortcut("Subscript"),
         sup: shortcut("Superscript"),
         strikethrough: "Strikethrough",
+        stack_snippet: {
```
threefjefff, line 46: Hmm, just realized that localization isn't something I've ported across. Is this even something we provide access to?

giamir: I think this does not belong in here anymore. My suggestion would be hardcoding it in the `stackSnippetsPlugin.ts` file directly and after having clarified the importance of i18n with business for the editor (and snippets) potentially find out a way for plugins to be internationalized. It could be something that hooks into what we already have in this file or something different. 
What's important (at least to me) is that the actual default text should live alongside the plugin and not in the core part of the editor. 
What do you think?

threefjefff: Yeah, sounds do-able. I think as a first pass it's not necessary, we can put it as a subtask to work i8n back in. That way we can explore an external plugin interface for it, or hooking the whole editor into the existing moonspeak implementation or something (maybe that's a whole plugin in it's own right.)
Anyways, out of scope for this, for sure.

plugins/official/stack-snippets/src/common.ts
```diff
@@ -0,0 +1,294 @@
+import { Node as ProsemirrorNode } from "prosemirror-model";
+import { generateRandomId } from "../../../../src/shared/utils";
```
giamir, line 2: I think we should keep references to editor source code to a minimum in plugins and make those references explicit. I would advocate for exposing the module you need in the `src/index.ts` and then I believe you should be even able to reference them here using the package name `import { module } from "@stackoverflow/stacks-editor"`. Basically I am advocating to be much more explicit about the core parts of the editor that are exposed for plugins to use (intentional coupling). If the package reference does not work I would also be ok to use a relative reference to `src/index` for now but the idea is to have a clear layer where we understand what core parts are "public".

This will help us in the long run to evolve the codebase with confidence.

plugins/official/stack-snippets/src/commands.ts
```diff
@@ -0,0 +1,55 @@
+import { MenuCommand } from "../../../../src/shared/menu";
```
giamir, line 1: See "I think we should keep references..." comment.

plugins/official/stack-snippets/src/paste-handler.ts
```diff
@@ -0,0 +1,192 @@
+import { Plugin, NodeSelection } from "prosemirror-state";
+import { EditorView } from "prosemirror-view";
+import { Node as ProsemirrorNode, Schema, Slice } from "prosemirror-model";
+import { log } from "../../../../src/shared/logger";
```
giamir, line 4: See "I think we should keep references..." comment.

plugins/official/stack-snippets/src/paste-handler.ts
```diff
@@ -0,0 +1,192 @@
...
+import {
+    BeginMetaLine,
+    EndMetaLine,
+    LangMetaLine,
+    mapMetaLine,
+    validSnippetRegex,
+} from "./common";
+import { generateRandomId } from "../../../../src/shared/utils";
```
giamir, line 12: See "I think we should keep references..." comment.

plugins/official/stack-snippets/src/paste-handler.ts
```diff
@@ -0,0 +1,192 @@
...
+import { insertParagraphIfAtDocEnd } from "../../../../src/rich-text/commands/helpers";
```
giamir, line 13: See "I think we should keep references..." comment.

plugins/official/stack-snippets/src/schema.ts
```diff
@@ -0,0 +1,218 @@
+import MarkdownIt, { Token } from "markdown-it";
+import { MarkdownParser } from "prosemirror-markdown";
+import { MarkdownSerializerNodes } from "../../../../src/shared/markdown-serializer";
```
giamir, line 3: See "I think we should keep references..." comment.

plugins/official/stack-snippets/src/schema.ts
```diff
@@ -0,0 +1,218 @@
...
+import {
+    assertAttrValue,
+    getSnippetMetadata,
+    mapMetaLine,
+    RawContext,
+    validateMetaLines,
+    validSnippetRegex,
+} from "./common";
+import { Node as ProseMirrorNode, NodeSpec } from "prosemirror-model";
+import { generateRandomId } from "../../../../src/shared/utils";
```
giamir, line 13: See "I think we should keep references..." comment.

plugins/official/stack-snippets/src/snippet-view.ts
```diff
@@ -0,0 +1,155 @@
+import { Node as ProseMirrorNode } from "prosemirror-model";
+import { EditorView, NodeView } from "prosemirror-view";
+import {
+    getSnippetMetadata,
+    SnippetMetadata,
+    StackSnippetOptions,
+} from "./common";
+import { error } from "../../../../src/shared/logger";
```
giamir, line 8: See "I think we should keep references..." comment.

plugins/official/stack-snippets/src/stackSnippetPlugin.ts
```diff
@@ -0,0 +1,83 @@
+/** This is an effort to turn the existing structure back into an external plugin while keeping the same functionality*/
+import type { EditorPlugin } from "../../../../src";
```
giamir, line 2: This is good :)

plugins/official/stack-snippets/src/stackSnippetPlugin.ts
```diff
@@ -0,0 +1,83 @@
...
+    stackSnippetMarkdownSerializer,
+    stackSnippetPlugin as markdownPlugin,
+    stackSnippetRichTextNodeSpec,
+} from "./schema";
+import { Node as ProseMirrorNode } from "prosemirror-model";
+import { EditorView } from "prosemirror-view";
+import { StackSnippetView } from "./snippet-view";
+import { StackSnippetOptions } from "./common";
+import { stackSnippetPasteHandler } from "./paste-handler";
+import { makeMenuButton } from "../../../../src/shared/menu";
```
giamir, line 14: See "I think we should keep references..." comment.

plugins/official/stack-snippets/src/stackSnippetPlugin.ts
```diff
@@ -0,0 +1,83 @@
...
+import { getShortcut } from "../../../../src/shared/utils";
```
giamir, line 15: See "I think we should keep references..." comment.

plugins/official/stack-snippets/test/menu.test.ts
```diff
@@ -0,0 +1,24 @@
+import { createMenuEntries, MenuBlock } from "../../../../src/shared/menu";
```
giamir, line 1: See "I think we should keep references..." comment.

plugins/official/stack-snippets/test/render.test.ts
```diff
@@ -0,0 +1,115 @@
+import {
+    invalidSnippetRenderCases,
+    snippetExternalProvider,
+    validSnippetRenderCases,
+} from "./stack-snippet-helpers";
+import { RichTextEditor } from "../../../../src/rich-text/editor";
```
giamir, line 6: See "I think we should keep references..." comment.

plugins/official/stack-snippets/test/stack-snippet-helpers.ts
```diff
@@ -0,0 +1,169 @@
+import { stackSnippetPlugin } from "../src/stackSnippetPlugin";
+import { ExternalPluginProvider } from "../../../../src/shared/editor-plugin";
```
giamir, line 2: See "I think we should keep references..." comment.

plugins/official/stack-snippets/test/stack-snippet-helpers.ts
```diff
@@ -0,0 +1,169 @@
...
+import { StackSnippetOptions } from "../src/common";
+import { richTextSchemaSpec } from "../../../../src/rich-text/schema";
```
giamir, line 4: See "I think we should keep references..." comment.

plugins/official/stack-snippets/test/stack-snippet-helpers.ts
```diff
@@ -0,0 +1,169 @@
...
+import { Schema } from "prosemirror-model";
+import { MenuBlock } from "../../../../src/shared/menu";
```
giamir, line 6: See "I think we should keep references..." comment.

giamir (approved):
Thanks @threefjefff for addressing my comments. Let's merge it! 🙂

PR #401, opened 2025-03-20
feat: devx - add parser logging for each rule stage

threefjefff:
Adds the ability to configure logging for the parser stages. This allows development of markdown-it parsers, including debugging of rules (e.g. "Why is my new rule not being executed?") and unexpected states (e.g. "Why is my rule doing nothing with the current token state?")

- Core rules do not run silently, and are each always run; so we log the token stream after each rule.
- Block and Inline rules log before each rule is executed, as the rule chain stops being executed when a rule successfully transforms the block (e.g. returns true when not in silent mode)


giamir (commented):
@threefjefff Is there a way to inject these logging statements instead of hardcoding them in the source code like this?
It would be nice if the production bundle would not contain all this extra code for debugging.

I guess where I am getting at is if there is a way to get to the `defaultMarkdownItInstance` from the outside world. 🙂

threefjefff:
> @threefjefff Is there a way to inject these logging statements instead of hardcoding them in the source code like this? It would be nice if the production bundle would not contain all this extra code for debugging.
> 
> I guess where I am getting at is if there is a way to get to the `defaultMarkdownItInstance` from the outside world. 🙂

Great question. I wonder if this could be rewritten as an external plugin. Lemme take a stab at it.

threefjefff:
> @threefjefff Is there a way to inject these logging statements instead of hardcoding them in the source code like this? It would be nice if the production bundle would not contain all this extra code for debugging.

Turns out: yes, easily 😂

giamir (commented):
Thanks @threefjefff for porting this to a plugin. I really like it.
I have only left few minor comments, after those are addressed I am ok with merging.

giamir:
I just noticed there are some e2e tests failures, probably we should keep the plugin off by default for those tests.

giamir (approved):
Thank you

Issue #111, opened 2022-04-20
Official external plugin support


b-kelly:
**Is your feature request related to a problem? Please describe.**

Stacks-Editor requires external plugin support so consumers can add custom functionality and extend the supported markdown and rich text node support.

**Additional context**

This is required for supporting the various extra editor features across the Stack Exchange network, such as poker, go, chess and code snippets.

PR #408, opened 2025-03-27
feat(code-blocks): remove code block processors

alexwarren:
This is a tidy-up PR ahead of doing some work to add a language picker dropdown for codeblocks.

It removes the "code block processors" plugin feature. These could add an "Edit" toggle to the top right of a code block, and there was an example of using this to implement a Mermaid plugin.

However, that plugin was broken (the graph didn't render). There is no production usage of code block processors. There is a potential future requirement for something like these - see #111 which implemented the feature - but we might implement these a different way. In any case, we can always add this code back if we need to - for now it's adding clutter, so removing it simplifies things.


threefjefff (approved):
You love to see it. 🔥

giamir (approved):
Thanks @alexwarren for doing this cleanup in a separate PR. It will be easier later on to consult how we implemented `codeBlockProcessor` if needed.

PR #409, opened 2025-03-28
refactor(code-blocks): make code block language code clearer

alexwarren:
This is another PR that is a precursor to adding a language picker dropdown to codeblocks, In the interests of keeping things small and reviewable.

There are no functional changes in this PR, but I've rearranged various things to make the current behaviour a little more explicit.

Code changes:
- removed `codeblockOverrideLanguage` from the options, as this is not used. The intention of this setting was to provide a default fallback language, presumably from a question's tags on Stack Overflow. But we're not using it (and a question can have multiple tags, so arguably this is not a useful design)
- there is logic for caching the auto-detected language from HighlightJS. This was stored in a node's `language` attribute, but since it's _only_ for the auto-detected language (and not for when a user specifies one directly), I've renamed that to `autodectedLanguage` to make it clearer.
- in functions that return the language for a codeblock, made it explicit that the user-specified language takes precedence. Previously this worked "by accident" in a way, because the auto-detected language would be unset in this case.
- make it clearer that the previously-named `language` field on the code block element is the title for the language indicator, which has "(auto)" appended if it's an auto-detected language

### DIFF
--- a/src/rich-text/editor.ts
+++ b/src/rich-text/editor.ts
@@ -47,7 +47,6 @@ import { createMenuPlugin } from "../shared/menu/plugin";
 export interface RichTextOptions extends CommonViewOptions {
     /** Array of LinkPreviewProviders to handle specific link preview urls */
     linkPreviewProviders?: LinkPreviewProvider[];
-    codeblockOverrideLanguage?: string;
 }
 
 /*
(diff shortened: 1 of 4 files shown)


giamir (approved):
Thanks @alexwarren for the cleanup. I am good to see this merged. I will only ask you to coordinate with @threefjefff since he is touching similar part of the codebase in his PR (especially the override language bits)
https://github.com/StackExchange/Stacks-Editor/pull/400/files#diff-bf75a9b790dd127d490f20857482f94083622233536bbbad62eeca1cdf8f449dL50

threefjefff:
> Thanks @alexwarren for the cleanup. I am good to see this merged. I will only ask you to coordinate with @threefjefff since he is touching similar part of the codebase in his PR (especially the override language bits) https://github.com/StackExchange/Stacks-Editor/pull/400/files#diff-bf75a9b790dd127d490f20857482f94083622233536bbbad62eeca1cdf8f449dL50

I touch on it mostly as an easy means of testing a different feature. I can easily come at it a different way (and, in fact, one of the tests I've added even has a todo that highlights that the implementation has some weird holes in it), so I'm all for fixing the merge conflict on my side.

PR #412, opened 2025-04-01
feat: Run publish pipeline

threefjefff:
Adds a "Publish" step to the pipeline that should run on:
- Merge to main
- PRs with the `create-release` tag added

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -88,8 +88,13 @@ jobs:
             - name: Build and push package to npm
               env:
                   NODE_AUTH_TOKEN: ${{ secrets.NPM_API_KEY }}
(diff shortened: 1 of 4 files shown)


threefjefff:
From the current run, proof that releasing depends on testing, and skips if no tag is required:
image

threefjefff:
Currently the `npm publish` step fails, as it needs a secret added for the API key that the publish step will use. This'll need some action from someone with admin access to the repo (I'm assuming @dancormier or @giamir)

threefjefff:
Per the [stacks ADR](https://github.com/StackExchange/Stacks/pull/1576), we're moving to changesets. Better to start from scratch here, I think.

PR #413, opened 2025-04-01
Move to changesets for releases and publishing

threefjefff:
Migrates change tracking to [changesets](https://github.com/changesets/changesets) (as per [this ADR](https://github.com/StackExchange/Stacks/pull/1576/files)).

This uses the standard implementation for changesets, and the same readme copy as in other examples.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -1,0 +1,14 @@
+{
+  "$schema": "https://unpkg.com/@changesets/config@3.1.1/schema.json",
+  "changelog": [
+      "@changesets/changelog-github",
+      { "repo": "StackExchange/Stacks-Editor" }
+  ],
+  "commit": false,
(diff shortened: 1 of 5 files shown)


giamir (approved):
That looks good to me. Thanks @threefjefff 
I have added the service account `NPM_API_KEY` secret to the pipeline.

I think we should merge and see if the workflow is running correctly. It should not do much since we have not created any changesets yet but it should certainly not fail.

- After we observed that the workflow run we can make a commit directly in main adding few changesets of what has changed since v0.10.11 was released. ([history](https://github.com/StackExchange/Stacks-Editor/commits/main/)). When the changesets land in main the workflow should create a new release PR with changelog, etc...

- Merging that PR will take care of releasing (npm, github, etc..).

PR #414, opened 2025-04-01
Changeset catchup to latest release (should be 0.11.1)

threefjefff:
This is a seperate changeset release to track changes up from 0.10.11 -> (hopefully) 0.11.1.

Typically these files would be produced alongside changes, but we have only recently moved to changeset, and wanted a means of tracking this intialization seperate from the GH action workflow changes.

### DIFF
--- a/.changeset/four-carrots-remain.md
+++ b/.changeset/four-carrots-remain.md
@@ -1,0 +1,9 @@
+---
+"@stackoverflow/stacks-editor": minor
+---
+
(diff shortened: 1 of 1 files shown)


giamir (approved):
> from 0.10.11 -> (hopefully) 0.11.1.

I think it will be `0.11.0`

Looking good, happy to merge and see if the release PR gets created correctly.

PR #417, opened 2025-04-04
fix(rich-text): link editor selection highlight and prepopulate URL

alizaberger:
**Describe your changes**

This PR makes inserting links less confusing in the rich text view.

1. The text highlighted was losing its highlight after clicking “Insert Link” so it was confusing what was being linked. This PR adds a grey highlight to the selected text while the Link Editor is active. 
2. Prepopulates the URL field with `https://`

**PR Checklist**

**Environment(s) tested**

- Device: [e.g. desktop, iPhone6, etc.]

**Additional context**

Add any other context about the PR here.


src/rich-text/commands/index.ts
```diff
@@ -313,7 +313,7 @@ export function insertRichTextLinkCommand(
                 state.selection.content().content.firstChild?.textContent ??
                 null;
             const linkMatch = /^http(s)?:\/\/\S+$/.exec(selectedText);
-            linkUrl = linkMatch?.length > 0 ? linkMatch[0] : "";
+            linkUrl = linkMatch?.length > 0 ? linkMatch[0] : "https://";
```
dancormier, line 316: I'd find this a little less convenient because I'd paste a link here most often, so having the input prepopulated with `https://` adds a step of deleting the text. Instead, my suggestion is to use the [`placeholder` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/placeholder) with a value of `https://…` or something like "Add a link" similar to Google Docs:

image

I don't think the placeholder is entirely necessary as long as the Link URL field is focused upon invoking the link editor. 

All this said, I'd like to get others' opinions on this since it's entirely possible this is just a _me_ thing.

dancormier (commented):
I've taken a look at the code and tested the changes out. This is a nice improvement. The sort of little annoyance I've experienced dozens of times. Thanks for tackling this!

I only had one minor UI suggestion but I wouldn't let it hold up merging if you and/or others feel differently

src/rich-text/plugins/link-editor.ts
```diff
@@ -195,12 +198,15 @@ export class LinkEditor extends PluginInterfaceView<
...
         super.update(view);
+        const state = LINK_EDITOR_KEY.getState(view.state);
 
         if (this.isShown) {
-            const state = LINK_EDITOR_KEY.getState(view.state);
             if (state?.url) {
                 this.hrefInput.value = state.url;
-                this.validate(state.url);
+
+                if (state.url != "https://") {
```
giamir, line 207: NIT: Consider using identity (===) instead of equality (==) here. In JavaScript, == allows type coercion, which can lead to unexpected behavior—for example, 0 == '0' is true, but 0 === '0' is false. Using identity (===) ensures both the value and the type match, which tends to be safer and more predictable. 

99.9% of the time in JS we should use equality (===) to avoid surprises. Quirks of an old language written in 10 days. 🙂

src/rich-text/commands/index.ts
giamir: I agree with @dancormier. I don't think it is a good practice to prefill text input. Placeholders are a more conventional way of giving hints to users. I would recommend to check with your designer about this whole flow and seek their input. 🙂

giamir (changes requested):
Thanks, @alizaberger, for the PR! I noticed a small regression at the end of the flow—after the user clicks the save button, the popover with the link appears again, but the arrow is no longer pointing to the center of the link. See screenshots.

Before (see arrow centered)
Screenshot 2025-04-08 at 10 26 49

After (see arrow not centered anymore)
Screenshot 2025-04-08 at 10 27 41

I know @alexwarren did some work on links/popovers a while back, so you might want to rebase this branch. That might be all that’s needed to make the issue go away.

Apart from that, only minor things. Thanks for making the editor better. ❤️

alizaberger:
Whoops, did not realize auto-merge would work without an approval! Sorry!

giamir:
No problem, thanks for addressing the PR comments. 🙂

PR #429, opened 2025-04-22
feat(code-blocks): add language picker

alexwarren:
This PR adds a language picker to code blocks.

Create a code block, and there is a new drop-down icon to the right of the auto-detected language:

image

Click the language, or press `Ctrl+;` (PC) or `Cmd+;` (Mac) to open the language picker:

image

Start typing to filter down the list of languages:

image

Click a language, or select one with the arrow keys to set it. Or keep typing if the language is not listed, and press Enter.

The language of the code block is updated, and the relevant syntax highlighting is applied:

image

In Markdown view, the language has been set:

image

For demo purposes there is a hard-coded list of languages. For production use on Stack Overflow, we'll inject the list of languages as part of the highlighting options when instantiating the editor.


src/rich-text/node-views/code-block.ts
```diff
@@ -27,29 +91,39 @@ export class CodeBlockView implements NodeView {
             return false;
         }
 
-        const newLanguageDisplayName = this.getLanguageDisplayName(node);
+        this.node = node;
+
+        this.dom.querySelector(".js-language-indicator").textContent =
```
threefjefff, line 96: Can this be moved after `const input` and then called as `input.textContent =`?

test/rich-text/node-views/code-block.test.ts
```diff
@@ -35,3 +53,144 @@ console.log("Hello World");
...
+    let richText: RichTextEditor;
+
+    beforeEach(() => {
+        richText = new RichTextEditor(
+            document.createElement("div"),
+            "",
+            externalPluginProvider([testCodeBlockPlugin])
+        );
+        // seed with a JS codeblock
+        richText.content = `~~~js
```
threefjefff, line 67: I'd like to see a test case for the other kind of code block (represented in markdown as four-spaces rather than these gates). Two things I'm concerned about are converting them to gates if a language gets specifed, and not converting them to gates if blured and not committed.

src/rich-text/node-views/code-block.ts
alexwarren: No, that's a different element (`.js-language-input`, whereas this is `.js-language-indicator`)

test/rich-text/node-views/code-block.test.ts
alexwarren: I've added this now in the tests (after an extensive back and forth with ChatGPT), just need to wire the new serializer plugin up to the real code now (which is frying my brain so I'll look at this tomorrow).

src/rich-text/node-views/code-block.ts
```diff
@@ -10,15 +10,79 @@ import { escapeHTML } from "../../shared/utils";
...
-    constructor(node: ProsemirrorNode) {
+    private render() {
         this.dom = document.createElement("div");
         this.dom.classList.add("ps-relative", "p0", "ws-normal", "ow-normal");
-        this.render();
+        this.dom.innerHTML = escapeHTML`
+        <button class="js-language-selector ps-absolute t2 r4 fs-caption fc-black-500 c-pointer baw0 bg-transparent" contenteditable="false">
+            <span class="js-language-indicator"></span>
+            <span class="svg-icon-bg iconArrowDownSm"></span>
+        </button>
```
dancormier, line 43: ## Suggestion: Use `.s-btn` classes for the button styling

Using Stacks' `.s-btn` classes will provide accessible focus and hover styles that aren't present currently (this issue especially apparent in dark mode). It will also allow us to get rid of the `<span class="svg-icon-bg iconArrowDownSm"></span>` element since the arrow would be provided by the `.s-btn__dropdown` class and should make this easier to maintain.

```diff
- <button class="js-language-selector ps-absolute t2 r4 fs-caption fc-black-500 c-pointer baw0 bg-transparent" contenteditable="false">
+ <button class="s-btn s-btn__muted s-btn__dropdown s-btn__xs ps-absolute t2 r4 js-language-selector" contenteditable="false"> 
```

dancormier, line 43: ## Suggestion: Add descriptive text to the button

It should be obvious to screen reader users what this button does before clicking it. As it stands, a screen reader would read the currently selected language. I suggest adding an element like this with some descriptive text for screen readers:

```diff
+ <span class="v-visible-sr">Change selected language. Currently: </span>
<span class="js-language-indicator"></span>
```

The wording could probably be cleaned up but it's roughly what I think is needed.

dancormier, line 43: ## Suggestion: Add `type="button"` to button element

The `type="button"` attribute should be set on each button. From [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/button_role#description):

> The button role identifies an element as a button to assistive technology such as screen readers. A button is a widget used to perform actions such as submitting a form, opening a dialog, canceling an action, or performing a command such as inserting a new record or displaying information. Adding `role="button"` tells assistive technology that the element is a button but provides no button functionality. Use [`<button>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button) or [`<input>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input) with `type="button"` instead.

src/rich-text/node-views/code-block.ts
```diff
@@ -10,15 +10,79 @@ import { escapeHTML } from "../../shared/utils";
...
+        <div class="ps-absolute t24 r4 js-language-input">
+            <div class="ps-relative mb8">
+                <label class="v-visible-sr" for="example-search">Search</label>
+                <input type="text" class="s-input s-input__search fs-caption js-language-input-textbox" placeholder="Search for a language" contenteditable="false" />
+                <span class="s-input-icon s-input-icon__search svg-icon-bg iconSearchSm"></span>
+            </div>
+            <div class="s-card fs-caption c-pointer py4 px4 js-language-dropdown-container">
+                <ul class="s-menu js-language-dropdown"></ul>
+            </div>
+        </div>
```
dancormier, line 53: ## Suggestion: Use `.s-popover` as the container for the language selector

> [!NOTE]
> FWIW, I'm not 100% on this suggestion. I'd suggest running it by a designer.

It seems we've added some custom code to show/hide the language selector when we could use the functionality built into Stacks by instead using the [`.s-popover` component](https://stackoverflow.design/product/components/popovers/). This would allow us to simplify the functional code in this file and get the benefits that come from using the popover (things like automatic layout). This would potentially change the design of the language selector unless we add a bunch of atomic styling classes to the popover, but there could be a styling middle-ground that allows us to use the `.s-popover` element

src/rich-text/node-views/code-block.ts
```diff
@@ -10,15 +10,79 @@ import { escapeHTML } from "../../shared/utils";
...
-        this.render();
+        this.dom.innerHTML = escapeHTML`
+        <button class="js-language-selector ps-absolute t2 r4 fs-caption fc-black-500 c-pointer baw0 bg-transparent" contenteditable="false">
+            <span class="js-language-indicator"></span>
+            <span class="svg-icon-bg iconArrowDownSm"></span>
+        </button>
+        <div class="ps-absolute t24 r4 js-language-input">
+            <div class="ps-relative mb8">
+                <label class="v-visible-sr" for="example-search">Search</label>
+                <input type="text" class="s-input s-input__search fs-caption js-language-input-textbox" placeholder="Search for a language" contenteditable="false" />
```
dancormier, line 47: ## Suggestion: Add appropriate `id` to `input`, update `label`

Currently, the input has no `id` and the `label` for this input has the attribute `for="example-search"`. I also think we should change the label text to be more descriptive. Something like this maybe:

```diff
- <label class="v-visible-sr" for="example-search">Search</label>
- <input type="text" class="s-input s-input__search fs-caption js-language-input-textbox" placeholder="Search for a language" contenteditable="false" />
+ <label class="v-visible-sr" for="code-block-language-input">Search languages</label>
+ <input id="code-block-language-input" class="s-input s-input__search fs-caption js-language-input-textbox" type="text" placeholder="Search for a language" contenteditable="false" />
```

dancormier (changes requested):
Nice enhancement @alexwarren. This is something I've wanted _personally_, so I'm selfishly happy to see this finally coming.

I've performed a _partial_ review and made some suggestions on the markup. Happy to discuss the suggestion further if you'd like a little Stacks/accessibility help.

I also noticed two bugs:

1. Any string can be set as the language




2. Languages not included in the predefined languages are set unexpectedly.

> [!NOTE]
> After reviewing, I realized this was mentioned in the description:
> > Or keep typing if the language is not listed, and press Enter.
>
> I'd consider this on that bug/feature line but I wouldn't cry if this stayed as-is.



These two bugs probably shouldn't be show-stoppers, but I figured they were worth mentioning.

test/rich-text/node-views/code-block.test.ts
alexwarren: I've implemented this "properly" now.

dancormier (approved):
Thanks for making the UI changes @alexwarren. They work well. I only found a couple of minor issues (one mentioned inline; the other here) that I don't consider blockers 🎉

#### Return creates new line unexpectedly

When pressing return when inputting an arbitrary string into the language selector, a new line is added to the editor. In this gif, I've removed most of what comes before the code block to demonstrate:

src/rich-text/node-views/code-block.ts
```diff
@@ -10,15 +10,79 @@ import { escapeHTML } from "../../shared/utils";
...
+        const textbox = this.dom.querySelector(".js-language-input-textbox");
+        textbox.addEventListener("blur", this.onLanguageInputBlur.bind(this));
+        textbox.addEventListener(
+            "keydown",
+            this.onLanguageInputKeyDown.bind(this)
+        );
+        textbox.addEventListener(
+            "mousedown",
+            this.onLanguageInputMouseDown.bind(this)
+        );
```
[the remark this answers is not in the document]
alexwarren: It prevents ProseMirror from erroring when you triple-click the textbox (as you might if you want to select all the text in it). Added a comment so this is clearer.

PR #430, opened 2025-04-25
feat(snippets): show/hide link functionality

alizaberger:
[Jira issue](https://stackoverflow.atlassian.net/browse/DISCOCD-81)

**Describe your changes**

Adds a "Show/Hide" link to toggle the visibility of code snippets (not the results pane) in the StackSnippetView. This corresponds to how snippets preview worked in the old editor. The link is only rendered when the hide attribute is "true". 

**PR Checklist**

**Environment(s) tested**

- Device: desktop
- Browser: chrome

**Additional context**

TODO: Need to add the same functionality to the preview pane in the MD+P mode.

### DIFF
--- a/plugins/official/stack-snippets/src/snippet-view.ts
+++ b/plugins/official/stack-snippets/src/snippet-view.ts
@@ -47,7 +47,7 @@ export class StackSnippetView implements NodeView {
             const toggleLink = document.createElement("a");
             toggleLink.href = "#";
             toggleLink.className = "snippet-toggle fs-body1";
-            toggleLink.textContent = codeIsShown ? "Hide Code" : "Show Code";
+            toggleLink.textContent = codeIsShown ? "Hide code snippet" : "Show code snippet";
             toggleContainer.appendChild(toggleLink);
 
             snippetContainer.appendChild(toggleContainer);
@@ -171,7 +171,7 @@ e
 
         const isVisible = node.attrs.showCode as boolean;
         snippetCode.style.display = isVisible ? "" : "none";
-        toggleLink.textContent = isVisible ? "Hide Code" : "Show Code";
+        toggleLink.textContent = isVisible ? "Hide code snippet" : "Show code snippet";
         arrowSpan.className = isVisible
             ? "svg-icon-bg iconArrowDownSm va-middle"
             : "svg-icon-bg iconArrowRightSm va-middle";
--- a/plugins/official/stack-snippets/src/snippet-view.ts
+++ b/plugins/official/stack-snippets/src/snippet-view.ts
@@ -166,19 +166,22 @@ export class StackSnippetView implements NodeView {
             JSON.stringify(this.snippetMetadata);
         this.snippetMetadata = updatedMeta;
 
-        // Update the visibility of the snippet-code div and toggle link
-        const snippetCode = this.contentDOM;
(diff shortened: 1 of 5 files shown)


plugins/official/stack-snippets/src/snippet-view.ts
```diff
@@ -19,19 +19,63 @@ export class StackSnippetView implements NodeView {
...
 
         //We want to render the language blocks in the middle of some content,
         // so we need to custom-render stuff here ("holes" must be last)
         const snippetContainer = document.createElement("div");
         this.dom = snippetContainer;
         snippetContainer.className = "snippet";
 
+        let toggleContainer: HTMLDivElement;
+
+        if (this.snippetMetadata.hide === "true") {
```
threefjefff, line 35: This matches current functionality. Out of scope for this PR, but I wonder if we could improve on what's here by always showing the show/hide accordian rather than keying it off of the snippet config (and instead setting the default based on that config value)

alizaberger:
@threefjefff Can you please take a look at [the failing test](https://github.com/StackExchange/Stacks-Editor/actions/runs/14657331881/job/41134607144?pr=430) and see if it has something to do with what I changed? I can't seem to figure it out

plugins/official/stack-snippets/src/snippet-view.ts
alizaberger: I also thought this behavior was interesting but I spoke to @KyleMit and it seems like users want to be able to control how the final post will look and if the snippet will have a show/hide link above it.

alizaberger:
> Pretty sure this is just a netlify thing, but it looks like the icon isn't being rendered properly in the preview there.
> 
> image
> 
> Functionality works great. Happy with the testing

No, this is happening on my local machine too, but the new snippets icon is acting the same for me locally. @dancormier any idea why the stacks icon is not rendering correctly?

plugins/official/stack-snippets/src/snippet-view.ts
KyleMit: I think what @threefjefff is still inline with my use cases (although agreed this matches existing behavior).  It wouldn't bug me if the link to be able to show/hide was always visible, but the default respected the value set by the config in the post.

That way OP has first say in whether it should be collapsed or visible, but users could also quickly collapse existing snippets in answers

dancormier:
@alizaberger I'm looking into the icon issue. In the meantime, can you provide instructions on how to test this functionality?

dancormier:
@alizaberger the editor relies on CSS definitions for icons (as opposed to importing SVGs), and only a small subset of icons are included in the icons CSS (see also https://github.com/StackExchange/Stacks-Icons/blob/production/README.md#using-the-css-icons).

To use this icon, I suggest importing the icon needed in [src/styles/icons.css](https://github.com/StackExchange/Stacks-Editor/blob/aberger/snippets-show-hide-link/src/styles/icons.css#L15) like this:

```css
…
.svg-icon-bg.iconArrowDownSm {
    --bg-icon: url("~@stackoverflow/stacks-icons/src/Icon/ArrowDownSm.svg");
}
…
```

dancormier (changes requested):
Thanks for the changes @alizaberger. I noticed a minor styling issue and have suggested some changes. Other than that minor issue, this is good to go from my perspective

dancormier (approved):

PR #436, opened 2025-04-30
fix(preview): Preview showing in markdown-only mode

alizaberger:
**Describe your changes**

There was a strange bug occurring where if a user refreshed the page when in markdown + preview mode, and then switch first to rich-text mode and finally to markdown only mode, the preview was incorrectly showing. 

[screen-capture (4).webm](https://github.com/user-attachments/assets/a0b7b586-b829-46e1-8ba1-a92300d06823)

What was happening was that the user's preference was (correctly) setting the default for preview to be true, but the preview visibility toggle was not being triggered when changing from rich-text mode to markdown mode, since they both have preview set to false. 

This PR fixes the behavior by always setting the preview visibility when changing editor modes. 

**PR Checklist**

**Environment(s) tested**

- Device: desktop
- Browser: chrome

**Additional context**

[Jira](https://stackoverflow.atlassian.net/browse/DISCOCD-83)


.changeset/moody-rabbits-argue.md
```diff
@@ -0,0 +1,5 @@
+---
+"@stackoverflow/stacks-editor": minor
```
alexwarren, line 2: I don't think we need to bump the minor version for this, as it's a bugfix?

```suggestion
"@stackoverflow/stacks-editor": patch
```

threefjefff (approved):

PR #437, opened 2025-04-30
fix(snippets): Fix unescapableCommand handling, swallow commands for snippets

threefjefff:
**Describe your changes**

- Fix unescapableCommand triggering on every arrow down press within the Snippet (e.g. in the a central language block, pressing down would cause a blank space despite not being at the end of the Snippet)
- Run unescapableCommand checks on up arrows too (e.g. you should be able to move above a snippet with the cursors)
- Allow External Plugins commands to override pre-definied editor commands
  - Use the ability to override editor commands to swallow commands that are nonsensical for Snippets, and cause breaking changes to happen (e.g. inserting a line-rule in the middle of the Snippet)

### DIFF
--- a/plugins/official/stack-snippets/test/commands.test.ts
+++ b/plugins/official/stack-snippets/test/commands.test.ts
@@ -1,7 +1,7 @@
 import { Node } from "prosemirror-model";
 import { EditorState } from "prosemirror-state";
 import { SnippetMetadata, StackSnippetOptions } from "../src/common";
-import { openSnippetModal, swallowedCommandList } from "../src/commands";
+import { openSnippetModal } from "../src/commands";
 import { RichTextHelpers } from "../../../../test";
 import {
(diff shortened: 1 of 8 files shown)


plugins/official/stack-snippets/test/commands.test.ts
```diff
@@ -225,4 +226,41 @@ describe("commands", () => {
...
+            const event = new KeyboardEvent("keydown", {
+                ctrlKey: true,
+                key: "Enter",
+            });
+
+            view.editorView.someProp("handleKeyDown", (f) =>
+                f(view.editorView, event)
+            );
+
+            //The Dom is exactly the same - no change has occured
```
alizaberger, line 262:
```suggestion
            //The Dom is not the same - a change has occured
```

alizaberger (approved):
Looks great and works well!

PR #440, opened 2025-05-12
Add Result controls (Fullscreen, hide-results, edit snippet) buttons to Snippet View

threefjefff:
**Describe your changes**

Adds:
- An additional way to trigger the Edit Snippet command (an edit button on the snippet)
- Expand Snippet controll (and hide snippet control that returns to original view)
- Hide/Show Results button that... hides and shows the results!

### DIFF
--- a/plugins/official/stack-snippets/snippets.css
+++ b/plugins/official/stack-snippets/snippets.css
@@ -1,0 +1,22 @@
+.snippet-fullscreen {
+    position: fixed;
+    left: 1%;
+    top: 1%;
+    z-index: 999999;
+    background-color: var(--white);
+    width: 100vw;
+    height: 100vh;
+}
+
+.snippet-buttons {
+    margin-bottom: 0 !important;
+}
+
+.snippet-fullscreen-controls {
+    display: flex;
+}
+
+.snippet-result-buttons {
+    display: flex;
+    margin-bottom: 0 !important;
+}
--- a/plugins/official/stack-snippets/src/snippet-view.ts
+++ b/plugins/official/stack-snippets/src/snippet-view.ts
@@ -6,6 +6,7 @@ import {
     StackSnippetOptions,
 } from "./common";
 import { error } from "../../../../src";
+import { openSnippetModal } from "./commands";
 
 export class StackSnippetView implements NodeView {
     constructor(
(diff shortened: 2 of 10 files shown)


alizaberger (changes requested):
The code is exceptionally well-written and clear. I noticed 2 issues:

1. The expand button does not behave the way it did in the legacy editor and in the [Figma](https://www.figma.com/design/WjUGTRl1Q6agpgS4YnrrdF/Stacks-Editor-UI-Changes?node-id=2359-3801&t=hUAzeud5DmK2qVLA-0). There the behavior was to expand the entire snippet content, not just the results pane. The expanded snippet should have the snippet code, the buttons, and the results pane.
2. When building locally, I got his error. `Module not found: Error: Can't resolve '@stackoverflow/stacks-icons/src/Icon/StackSnippets.svg' in 'C:\Users\aberger\source\repos\Stacks-Editor\src\styles'`

plugins/official/stack-snippets/snippets.css
```diff
@@ -0,0 +1,22 @@
...
+.snippet-buttons {
+    margin-bottom: 0 !important;
+}
+
+.snippet-fullscreen-controls {
+    display: flex;
+}
+
+.snippet-result-buttons {
+    display: flex;
```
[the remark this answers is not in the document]
giamir: Thanks @threefjefff for the PR, I also agree with this comment. 

In case we are not able to use Stacks atomic classes for some reason and need custom syles we should bundle the styles in a separate file that need to be imported explicitly by consumers. In fact if there are only a bunch of styles I would recommend to go inline rather than create a bundle so that the consumers don't need to import 2 bundles (JS and CSS) to make the plugin work. 

So here is my recommendation:
- Try to achieve your styling need with stacks atomic classes first
- (if that does not work) apply the styles directly where you need them (inline styling)
- (last resort - hopefully we don't need this) create a dedicated css bundle for the plugin

Let's try not to mix plugin styles with editor core styles the way this PR is doing at the moment (see webpack config changes that were made). 

I can see we are also adding snippet related icons to the central icons.css file. 
This also violates Core/Plugin boundaries. If an icon is exclusively needed for a plugin it should not be added in a Core editor bundle but instead stay close to the plugin. We could probably also inline those icons in the plugin javascript (untested pseudo code):
```ts
const shareSmUrl = new URL('@stackoverflow/stacks-icons/src/Icon/ShareSm.svg', import.meta.url);
const iconShareSm = document.querySelector('.svg-icon-bg.iconShareSm');
iconShareSm.style.width = '18px';
iconShareSm.style.setProperty('--bg-icon', `url(${shareSmUrl})`);
```
I am personally ok to be pragmatic and keep the icons in the central file for now but I would appreciate if we could at least add a comment specifying the icons that are needed only for the stacks-snippets plugin so that in future we could find a solution that doesn't violate encapsulation.

Thanks again and happy to chat sync about this if needed.

giamir (changes requested):
Happy to approve as soon as the styling part is addressed. Thanks @threefjefff

plugins/official/stack-snippets/snippets.css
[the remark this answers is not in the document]
threefjefff: The styles were there to pull together a bunch of the atomic classes that would need added/removed at once. I've kept the class name and since check for it's presence to add/remove the atomic classes as we need to.

As for the icons, the whole file is annotated with:
> /* TODO get rid of these once they're exposed in the upstream Stacks-Icons package */

Frankly, loading the icons directly from the package that way via URL does not seem like a clean way of doing this, and it seems out-of-scope for what this PR is trying to do, so I've taken your advice to add a comment and leave those styles in for now.

plugins/official/stack-snippets/src/snippet-view.ts
```diff
@@ -184,7 +158,71 @@ export class StackSnippetView implements NodeView {
...
+            if (!this.resultContainer.classList.contains("d-none")) {
+                this.resultContainer.classList.add("d-none");
+            }
+        }
+
+        //Fullscreen the results, if the node meta needs it
+        if (content && node.attrs.fullscreen) {
+            if (!this.dom.classList.contains("snippet-fullscreen")) {
+                //We use `.snippet-fullscreen` as a marker for the rest of the styling
+                this.dom.classList.add("snippet-fullscreen");
```
giamir, line 189: NIT: I think you can simplify a bit this code, by passing more than one class at once
```
this.dom.classList.add("class1", "class2", etc...
```
Same for remove. You could even put them in an array and spread it. 
https://developer.mozilla.org/en-US/docs/Web/API/Element/classList#examples

giamir (approved):
Thanks for addressing my comment @threefjefff

src/styles/icons.css
```diff
@@ -27,3 +27,17 @@
 .svg-icon-bg.iconStackSnippets {
     --bg-icon: url("~@stackoverflow/stacks-icons/src/Icon/StackSnippets.svg");
 }
+
+/* Required for the plugin/official/stackSnippet plugin */
+.svg-icon-bg.iconShareSm {
+    width: 18px;
```
alizaberger, line 33: In the figma I think this is the original size, 14px.

plugins/official/stack-snippets/src/snippet-view.ts
```diff
@@ -203,12 +231,208 @@ export class StackSnippetView implements NodeView {
...
+        container.appendChild(expandButton);
+        return expandButton;
+    }
+
+    private buildFullscreenCollapseButton(
+        container: HTMLDivElement
+    ): HTMLButtonElement {
+        const collapseButton = document.createElement("button");
+        collapseButton.type = "button";
+        collapseButton.className = "s-btn flex--item td-underline ml-auto";
```
alizaberger, line 422: The figma uses the same icon here as the expand button

plugins/official/stack-snippets/src/snippet-view.ts
```diff
@@ -83,73 +85,38 @@ export class StackSnippetView implements NodeView {
...
+            snippetResultButtonContainer.className =
+                "snippet-result-buttons d-flex mb0 d-none ml-auto gs4";
+            ctas.appendChild(snippetResultButtonContainer);
+            this.showButton = this.buildShowButton(
+                snippetResultButtonContainer
+            );
+            this.hideButton = this.buildHideButton(
+                snippetResultButtonContainer
+            );
+            this.fullscreenButton = this.buildFullscreenExpandButton(
```
alizaberger, line 106: The fullscreen expand/collapse buttons should be visible even when the snippet has not been run yet, i.e. not dependent on the results container

alizaberger (changes requested):

src/styles/icons.css
threefjefff: Good call. Think it also had some extra spacing, but I'm struggling to find that (a lot of the pictures are screenshots). Where in the Figma are you pulling the measures from?

plugins/official/stack-snippets/src/snippet-view.ts
threefjefff: True, also the copy was wrong. Latest: 
image

plugins/official/stack-snippets/src/snippet-view.ts
threefjefff: Ah, good catch. This is lingering from before, where it was expanding just the content. Updated the behviour now.

alizaberger (approved):

PR #457, opened 2025-06-17
fix(stack-snippet): expose stack-snippet options and metadata types

troygould-stack:
**Describe your changes**

A couple of options interfaces passed to the StacksEditor are not exported. Therefore, I had to recreate these interfaces when I wanted to use them. See: 
https://github.com/StackEng/StackOverflow/blob/571e63834ee5920c9f5db16b97b84f94c85a872d/StackOverflow/_Scripts/islands/ask-question/src/shared/stackSnippets.ts#L18

### DIFF
--- a/plugins/official/index.ts
+++ b/plugins/official/index.ts
@@ -1,2 +1,5 @@
 export { stackSnippetPlugin } from "./stack-snippets/src/stackSnippetPlugin";
-export type { StackSnippetOptions, SnippetMetadata } from "./stack-snippets/src/common";
+export type {
+    StackSnippetOptions,
+    SnippetMetadata,
+} from "./stack-snippets/src/common";
--- a/plugins/official/index.ts
+++ b/plugins/official/index.ts
@@ -1,1 +1,3 @@
 export { stackSnippetPlugin } from "./stack-snippets/src/stackSnippetPlugin";
+export { StackSnippetOptions } from "./stack-snippets/src/common";
+export { SnippetMetadata } from "./stack-snippets/src/common";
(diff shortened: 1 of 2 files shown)


plugins/official/index.ts
```diff
@@ -1 +1,3 @@
 export { stackSnippetPlugin } from "./stack-snippets/src/stackSnippetPlugin";
+export { StackSnippetOptions } from "./stack-snippets/src/common";
+export { SnippetMetadata } from "./stack-snippets/src/common";
```
giamir, line 3: NIT: make it explicit we are exposing types
```suggestion
export type { StackSnippetOptions, SnippetMetadata } from "./stack-snippets/src/common";
```

giamir (commented):
@troygould-stack Thanks for the PR.

These changes need a patch release in order to be used in Core. To do that you can use `npx @changeset/cli` in this PR branch to generate a changeset. Once the PR is merged you can publish the package in autonomy by following the instructions in [this part](https://github.com/StackExchange/Stacks-Editor?tab=readme-ov-file#publishing) of the README. 

_Sidenote: Instead of hardcoding things in Core an alternative workaround would have been to write a reference to the internals of the editor I believe (I am not 100% sure though and have not tested it):_
```
import type { SnippetMetadata } from "../../../../node_modules/@stackoverflow/stacks-editor/dist/plugins/official/stack-snippets/src/common";
```
_This PR is the right call, you don't want to dig into internals of packages like that in general since internals should stay private. That said it might save you some copy/paste next time you need a temporary workaround._

giamir (approved):

PR #507, opened 2026-08-05
fix: preserve editor focus when opening dropdowns

alexwarren:
## Summary

- cancel `mousedown` within the built-in editor menu container so mouse activation does not blur the editor
- keep plugin controls outside the menu focusable
- add unit and browser regression coverage, with the test dependency updated to Stacks 2.9.0
- add a patch changeset

## Context

This fixes the underlying Stacks Editor focus-handling gap exposed by Stacks 2.9.0. When a mouse opened an editor dropdown, the editor click handler restored focus to the editor; Stacks then treated that focus move as leaving the trigger/popover pair and immediately dismissed the menu.

Stacks Editor already documents this focus-preservation behavior in two places: `setupPluginContainer()` says toolbar mouse interactions are cancelled to prevent the editor from blurring, and `MenuView` says the standard `StacksEditor` setup handles this automatically. However, the built-in setup never installed the documented `mousedown` handler. This change closes that existing implementation gap; Stacks 2.9's focus-leave behavior merely made the gap observable.

This is the upstream fix for the temporary Stack Overflow workarounds in [StackEng/StackOverflow#26004](https://github.com/StackEng/StackOverflow/pull/26004) and [StackEng/StackOverflow#26010](https://github.com/StackEng/StackOverflow/pull/26010).

## Testing

- `npm run lint`
- `npm run test:unit -- --runInBand` — 48 suites, 1,092 tests passed
- `npm run test:e2e` — 195 tests passed across Chromium, Firefox, and WebKit
- `npm run build`
- `git diff --check`

### DIFF
--- a/.changeset/quiet-menus-stay.md
+++ b/.changeset/quiet-menus-stay.md
@@ -1,0 +1,5 @@
+---
+"@stackoverflow/stacks-editor": patch
+---
+
+Keep editor dropdown menus open when activated with a mouse.
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,7 +37,7 @@
                 "@playwright/test": "^1.42.1",
                 "@stackoverflow/commitlint-config": "^1.0.0",
                 "@stackoverflow/prettier-config": "^1.0.0",
-                "@stackoverflow/stacks": "^2.3.2",
+                "@stackoverflow/stacks": "^2.9.0",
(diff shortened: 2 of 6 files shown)


dancormier (approved):
Really solid change @alexwarren. I looked through the code plus tested manually and it all seems good. Thanks!

